Issue #83, opened 2026-03-05
Service Canceling


sgunes16:
Implement and document the UX and edge cases for **service canceling** so users can reliably cancel agreements or remove their listings when allowed (handshake cancel, cancel own service listing).

## Background

- **Backend**:
  - **Handshake cancel**: `POST /api/handshakes/{id}/cancel/` — provider cancels an *accepted* handshake (refund, notification, service reverts to Active if it was Agreed).
  - **Service cancel** (remove listing): `DELETE /api/services/{id}/` — only owner (or admin), and only when the service has **no handshakes** and has **no agreed handshakes**.
- **Frontend**: Handshake cancel is available from Chat; Event cancel exists on service detail. Service cancel is exposed in `serviceAPI.delete()` but there is no clear UI for “Cancel service” / remove listing (e.g. on service detail or “My services”).

## Goals

1. **Handshake cancel**  
   Keep current behavior (e.g. from Chat); ensure copy and error handling are clear (e.g. “Only accepted handshakes can be cancelled”, “Only the service provider can cancel”).

2. **Service cancel (remove listing)**  
   Add a “Cancel service” action where it makes sense (e.g. service detail for owner, or “My services”). Show a clear error when cancel is disallowed because handshakes exist (e.g. “You can’t cancel this service because it has at least one agreed handshake. Cancel or complete those first.”).

Issue #76, opened 2026-03-04
Event chat error


sgunes16:
There are some error at event chat system and wrong setup for check in and handshakes. Fix is needed.

PR #86, opened 2026-03-05
feat: service cancel UX + event group chat fix (#83, #76)

yusufizzetmurat:
Service Canceling (Issue #83):
- Convert service DELETE to soft-delete (status=Cancelled)
- Only block removal when active handshakes exist
- Allow removal when only terminal handshakes remain
- Filter cancelled services from UserProfile for non-admins
- Add Remove Listing button on ServiceDetailPage and DashboardPage
- Improve handshake cancel error copy in ChatPage
- Add/update 5 backend tests for soft-delete behavior

Event Group Chat (PR #76 fix):
- Fix GroupChatViewSet to allow Event type services
- Fix GroupChatConsumer WebSocket access for events
- Accept checked_in and attended statuses for event chat access
- Add Event Chat button on ServiceDetailPage for organizer and participants
- Update ChatPage group eligibility to include events
- Add ?group={serviceId} deep-link for navigating to event group chat
- Add 4 backend tests for event group chat access


mzyavuz (approved):
The user cannot cancel when there is a handshake (cancel is not available), The user can cancel when there is no handshake, 
- thats my design choice, if there is an ongoing handshake user should not be able to remove that listing first user will need to cancel or complete handshake.

The event can be cancelled, when there is less than 24 hours to start it warn,

Group chats are great work, I love it but the information of the event is not given at the group chats (the customer may not like it)
- I need more clarification on this, event chats are just for basic event participants commınication

PR #327, opened 2026-03-29
feat: implement xfail tests for QR and GPS check-in validation

yusufizzetmurat:
- Added unit tests for QR token validation and GPS proximity check-in as part of FR-19i.
- Tests include scenarios for valid and invalid QR tokens, single-use token enforcement, and GPS coordinate validation.
- Marked tests as xfail due to incomplete implementation in EventHandshakeService.
- Enhanced test coverage for in-person and online event check-in requirements.


mzyavuz (approved):
Good structure — the green/xfail split is clear and the module docstrings explain exactly what is and isn't implemented. I'm approving.

PR #355, opened 2026-04-05
feat(tag-hierarchy): add parent_qid, entity_type, and depth fields to Tag model

### LINKED ISSUE #286: [FR-12b, FR-12c, FR-E01a, FR-E02a, FR-SEA-01] The tag system needs to be rethought, WikiData hierarchy is not being used
The spec requires categories on offers (FR-E01a), requests (FR-E02a), and events (FR-12b, FR-12c), and also requires WikiData-weighted semantic search (FR-SEA-01). All three of those requirements touch the same underlying problem: our current tag system is too flat to be useful.

**What we have right now:**

The `Tag` model stores only a WikiData QID (`id`) and a display `name`. There is no parent/child relationship, no entity type, no depth, no hierarchy at all. Tags on offers, requests, and events all go into the same flat table. The WikiData autocomplete lets you attach essentially any WikiData entity as a tag, Python the programming language" and "Python the snake" and "Python the comedy group" are all equally valid, and the system can't tell them apart.

This matters for the following reasons:

- WikiData entities have very different levels of abstraction. "Computer science" (Q21198) is a top-level field. "Python" (Q28865) is a specific language. "Django" (Q290053) is a specific framework. Right now a search for "programming" won't find services tagged only with "Python" because there is no hierarchy to traverse upward.
- FR-SEA-01 requires search to weight Wikidata tag matches at 0.8 vs title matches at 1.0. The current `TextStrategy` and `TagStrategy` in `search_filters.py` apply equal weight to everything — there is no scoring, only filtering.
- The `WikidataTagAutocomplete` component doesn't filter by entity type, so a user creating a cooking event could attach a geographic entity or a historical figure as a tag and the system accepts it silently.
- The frontend `DashboardPage` does its own client-side tag name search with plain string matching, which is completely disconnected from the backend WikiData structure.

**What a better system would look like:**

WikiData entities have a `P31` (instance of) and `P279` (subclass of) property chain that gives them a position in a hierarchy. If we stored even one level of parent context alongside each tag — its immediate parent QID and entity type — we could:

1. Group tags into meaningful browse categories (e.g., any tag that is a subclass of "programming language" maps to a "Technology" browse category) without adding a hardcoded enum to every service.
2. Implement hierarchical search: a search for "Technology" finds services tagged "Python", "Django", "JavaScript" etc., because the system knows those are all technology-domain entities.
3. Filter the WikiData autocomplete by entity type (only allow skills, activities, domains — not places or people) so tag quality improves at input time.
4. Feed the WikiData hierarchy into the FR-SEA-01 weighted search: a direct tag match scores 0.8, a parent-category match scores lower, a text match on the tag name scores lower still.
5. Tags that are needed for user profile creation should also be evaluated accordingly.

**Affected files across the codebase:**

- `backend/api/models.py` — `Tag` model needs at minimum `parent_id` and `entity_type` fields; `Service.tags` M2M is used by all three service types
- `backend/api/wikidata.py` — currently fetches only `label` and `description`; needs to also fetch `P31`/`P279` parent chain
- `backend/api/serializers.py` — `TagSerializer` and the tag creation logic inside `ServiceSerializer.create()` (the section that calls `enrich_tag_with_wikidata`) need to store parent context when creating tags
- `backend/api/search_filters.py` — `TagStrategy` does a flat `tags__id__in` filter; `TextStrategy` searches tag names with equal weight; neither implements FR-SEA-01 scoring
- `backend/api/views.py` — `WikidataSearchView` proxies search with no entity-type filtering
- `frontend/src/components/WikidataTagAutocomplete.tsx` — no entity-type guard at input time
- `frontend/src/components/ServiceForm.tsx` — tag submission logic; no type restriction
- `frontend/src/pages/DashboardPage.tsx` — client-side tag name search should be removed in favour of the backend search once scoring is in place
- `frontend/src/services/tagAPI.ts` — `tagAPI.search()` and `tagAPI.searchLocal()` may need to pass entity-type hints

**Before writing any code, decide:**
- How many levels of parent hierarchy to store
- Which WikiData entity types are allowed as tags (skills/activities/domains only, or broader)
- Whether the browse category grouping comes from the stored parent or from a separate mapping table
- How aggressively to backfill existing tags with parent data

**Acceptance criteria:**
- The `Tag` model stores at least the immediate WikiData parent QID and a broad entity type classification.
- The WikiData autocomplete filters suggestions to skill/activity/domain entities and does not suggest places, people, or other irrelevant entity types.
- A search for a parent concept (e.g., "programming") surfaces services tagged with child concepts (e.g., "Python", "Django") via the hierarchy, not just by string match.
- `TagStrategy` and `TextStrategy` in `search_filters.py` implement the FR-SEA-01 weighted scoring (title 1.0, tag match 0.8) instead of flat filtering.
- Existing tags are backfilled with parent data without data loss; a migration is provided.
- The category grouping used in FR-12b/FR-12c for events and in FR-E01a/FR-E02a for offers and requests is derived from the tag hierarchy, not from a separate hardcoded field.
- Unit tests for the updated `TagStrategy` with hierarchy traversal; integration tests for category-grouped browse; E2E for the restricted autocomplete.

yusufizzetmurat:
**Summary**

  The tag system was flat: a Tag stored a QID and a name, no hierarchy. Searching "programming" couldn't find a service tagged "Python" because there was no parent relationship. The
  autocomplete accepted any WikiData entity including cities and people.

  This PR stores one level of WikiData hierarchy per tag (parent QID + entity type) and uses it to power hierarchical search, weighted scoring (FR-SEA-01), and entity-type filtering on
  autocomplete.

  **Backend:**
  - Tag model gains parent_qid, entity_type, depth fields (nullable, non-breaking)
  - wikidata.py: new functions to fetch P31/P279 claims, resolve entity types via a curated map of ~25 parent QIDs, and filter autocomplete results (blocks places, people, countries)
  - search_filters.py: TagStrategy traverses parent_qid for hierarchical matching. SearchEngine annotates a search_score (title=1.0, tag=0.8, description=0.6, parent=0.5, tag name=0.3) and
   orders by it
  - WikidataSearchView returns entity_type per result and filters out irrelevant types
  - New tags get hierarchy data on creation. Backfill command for existing tags

  **Frontend:**
  - Autocomplete shows entity type badges and descriptions per suggestion
  - DashboardPage delegates search to backend instead of client-side string matching
  - Tag interface extended with parent_qid, entity_type, description

  No changes to services.py or auth/handshake/reputation error handling in views.py, so zero conflict with epic/backend-refactor.

  **How to test**

  Run new tests (48 added, 501 total, all passing):
  cd SWE-574-3/backend && source .venv/bin/activate
  pytest api/tests/unit/ -v --no-cov --override-ini="addopts="

  Backfill existing tags:
  python manage.py backfill_tag_hierarchy --dry-run   # preview
  python manage.py backfill_tag_hierarchy              # run for real

  **Manual checks:**
  - Create service, type "London" in tag autocomplete: should not appear (city, filtered out)
  - Type "Python": should appear with a Technology badge and description
  - Create a service tagged "Python", then search "programming" on the dashboard: the service should appear via parent traversal
  - GET /api/services/?tag=Q9143 returns services tagged with child concepts (Python, Django, etc.)
  - Search "Python": service with "Python" in the title ranks above a service that only has the Python tag
  - cd frontend && npx tsc --noEmit compiles clean

Closes #286


sgunes16 (approved):
Works well but some enhancements can be added also. But this is a good starting point

PR #392, opened 2026-04-11
Firebase setup

### LINKED ISSUE #391: Set up Firebase Cloud Messaging for Push Notifications
Firebase Cloud Messaging is the first firebase module that is going to be integrated the to project so Firebase setup needs to be done in the first place. 

Acceptance Criteria:
1. Firebase Project is created.
2. Necessary npm packages are installed to the mobile client project.
3. google-services.json and GoogleService-Info.plist are added to the mobile client  project.
4. FCM setups done. 
5. A spec file for the BE to integrate sending push notifications is prepared. 
6. Notifications should include data inside to indicate where these push notifications will navigate to when they are opened.
7. Client side should implement navigation from pushes.

diclenaz7:


yusufizzetmurat (changes requested):
Reviewed the PR. Please don't merge yet, there are a few issues:

  1. Bundle ID mismatch: google-services.json and GoogleService-Info.plist use com.apiary.thehive but the app uses com.diclenaz.thehive. Firebase won't initialize, either change the app or Firebase.
  2. Duplicate foreground notifications: Both Expo's notification handler and FCMService's Alert.alert() will fire for the same message, so users see every notification twice.
  3. Permission prompt before login: requestUserPermission() runs unconditionally in App.tsx before auth. Unauthenticated users get permission dialogs.
  4. getFCMToken() is never called: The backend only uses Expo push tokens. The FCM token is never registered anywhere, it's dead code.
  5. Firebase credential files committed: google-services.json and GoogleService-Info.plist with API keys are tracked in git and not in .gitignore.

 The docs file looks good, however everything it describes is already implemented in the backend.

diclenaz7:
Thanks for the thorough review @yusufizzetmurat ! All five points have been addressed in this commit:

Bundle ID mismatch — Resolved by aligning everything on com.apiary.thehive. Both google-services.json (package_name) and GoogleService-Info.plist (BUNDLE_ID) now match the bundleIdentifier/package declared in app.json. Firebase will initialize correctly on both platforms.

Duplicate foreground notifications — Fixed by removing FCMService entirely. The subscribeForegroundMessages() call (which showed a redundant Alert.alert()) has been removed from App.tsx. Foreground notification display is now handled exclusively by usePushNotifications via Expo's setNotificationHandler, so each message surfaces exactly once.

Permission prompt before login — Resolved as a side effect of the above. The unconditional requestUserPermission() call in App.tsx is gone. Permission is now requested only inside usePushNotifications, which gates everything behind isAuthenticated, so unauthenticated users will never see a permission dialog.

Dead getFCMToken() code — FCMService.tsx has been deleted in full. Since the backend exclusively consumes Expo push tokens (registered/deregistered in usePushNotifications), there was no reason to keep the FCM token path around.

Firebase credentials in git — google-services.json and GoogleService-Info.plist are now listed in .gitignore. Template files (google-services.json.example, GoogleService-Info.plist.example) with placeholder values have been added so new contributors know the expected structure. Note: the real credential files were already committed historically — running git rm --cached google-services.json GoogleService-Info.plist will scrub them from the index without deleting them locally.

yusufizzetmurat:
Thanks for addressing the earlier feedback! Most issues are resolved, but there's still one blocker:

Firebase credentials are still tracked in git. Altough the latest commit fixes it we need to delete the history as it is exposed. 

Can you please rebase and squash it so that it is deleted from the commit history

yusufizzetmurat (approved):

PR #440, opened 2026-04-28
feat(browse): implement category filter and fix browse page UX issues

### LINKED ISSUE #172: Adding Event Color to the Map View (Yellow)
Events text and color should be added to the Map View, next to Wants

Image

### LINKED ISSUE #178: Move closed services at the list to the end
Currently, closed or declined services remain mixed within the service list, making it harder to separate active and inactive items. Update the logic so that all closed services—including those no longer available to the user because they have already declined—are displayed at the end of the list. This will improve usability by keeping active and available services more visible at the top.

### LINKED ISSUE #183: Started Events are Shown on Browse Page & Editable on Detail Page
**Problem:**
Events that have already started are still visible on the browse page. These started events can be edited via the frontend UI, but attempting to save changes results in backend errors.

**Where:**
- Browse Page (frontend)
- Event editing forms

**Expected Behavior:**
- Started events should not appear on the browse page, or at the very least, should not be editable if they have already started.
- If the frontend allows editing, the backend should either support it (if intended) or the option should be disabled completely.

**Steps to Reproduce:**
1. Go to the browse page where all events are listed.
2. Observe that events which have already started are still listed.
3. Click to edit one of these started events. Edit the details.
4. Click save and see that backend returns an error (since editing is not permitted for started events).

**Impact:**
Users may think they can update started events, leading to confusion and broken UX when they encounter backend errors.

**Suggestions:**
- Hide started events from the browse page or, alternatively, disable the edit option for them on the frontend.
- Ensure frontend and backend implementations are consistent about what operations are allowed for started events.

### LINKED ISSUE #186: Implement Filter at Browse page for Event, Offer, and Need
**Problem:**
Currently, users who are browsing events, offers, or needs have no way to filter the results on the Browse page. This makes it difficult to find relevant entries efficiently, especially as the number of events, offers, and needs grows.

**Expected Behavior:**
- The Browse page should provide filtering options so users can narrow down the displayed list by category (Event, Offer, Need).
- Additional filters may be added in the future (such as by date, location, etc.), but category-based filtering is the first priority.

**Suggestions:**
- Add a filter UI on the Browse page that allows users to select whether to view Events, Offers, Needs, or any combination thereof.
- Update the backend/API as necessary to support filtering requests.
- Clearly indicate to users which filters are currently active.

This will improve user experience and make it more efficient to find relevant content.

mzyavuz:
# Summary
Implements the category filter (Event / Offer / Need) on the Browse page and resolves several related UX and correctness issues discovered during development.

Closes #186
Closes #183
Closes #178
Closes #172                                                                                                

# Changes
  - Browse category filter — added Offers / Needs / Events toggle chips into the existing filter bar (both desktop pill  
  group and mobile scroll strip); chips are multi-select and compose with existing filters (New, Online, Recurrent,
  Weekend); colour-coded to match the map legend                                                                         
  - Map legend — added missing Events (amber) entry alongside Offers and Needs
  - Backend TypeStrategy — fixed to accept 'Event' as a valid type query param; was silently ignored before              
  - Started events hidden from Browse — backend queryset and client-side displayServices now exclude Events whose scheduled_time is in the past; isWithinLockdownWindow upper bound removed so the edit lock stays engaged after an event starts, matching backend behaviour.
  - Closed/declined services sorted to end — services with a denied or cancelled handshake are sorted to the bottom of   
  the feed via a stable sort, preserving existing hot-score/pin/recency order for all other entries                      
  -  Remove Listing confirmation — replaced window.confirm with AdminConfirmModal
  -  Cancel Event confirmation — replaced window.confirm with a styled modal; includes a required reason textarea
   and an amber lockdown warning when cancelling within 24 h with active participants; reason is stored on               
  service.cancellation_reason                                                                                            
  - Evaluation — opposite trait exclusivity — selecting a trait now deselects only its direct opposite (Punctual ↔ Late, 
  Helpful ↔ Unhelpful, Kind ↔ Rude, and event equivalents); unrelated cross-group combinations remain valid              
  - Evaluation — image upload — multiple attribute added so up to 3 images can be selected at once; files beyond
  available slots are rejected with a toast; files over 10 MB are rejected by name; HTTP 413 from the server shows a     
  clear size error; upload failure keeps the modal open so the user can remove images and retry without re-submitting the
   evaluation                                                                                                            
                                                            
# How to Test
  1. Go to the Browse page — confirm Offers, Needs, and Events chips appear inside the existing filter bar               
  2. Toggle each chip individually and in combination; verify the grid and map update accordingly
  3. Combine a category chip with an existing filter (e.g. Events + New); confirm both apply                             
  4. Find or create an Event whose scheduled_time is in the past — confirm it does not appear on Browse and that its edit
   button is disabled on the detail page                                                                                 
  5. On a listing where you have a denied or cancelled handshake, confirm it appears at the bottom of the feed           
  6. As a listing owner, click Remove Listing — a styled confirmation modal should appear; cancel dismisses it, confirm  
  removes it                                                                                                             
  7. As an event organiser, click Cancel Event — a modal should appear; the confirm button must be disabled until a      
  reason is typed; if within the 24 h lockdown with participants, the amber ban warning should be visible; submit and    
  verify the reason is saved                                
  8. Open the evaluation modal — select a positive trait, then its negative opposite; the positive trait should deselect;
   select a non-opposite cross-group pair (e.g. Late + Helpful) — both should remain selected                            
  9. In the evaluation modal, attempt to select 4 images at once — a toast should warn and only 3 should be added;
  attempt to add a file over 10 MB — a named error toast should appear and the file should be skipped                    
  10. Simulate a 413 response on image upload — the modal should stay open with images intact; remove one image and
  resubmit — the upload should retry without re-submitting the evaluation traits                                         
                                                            
# Screenshots
Screenshot 2026-04-28 at 15 48 51


yusufizzetmurat (dismissed):
Nice work on the bundle, the cancel/remove modals, the multi-image upload retry pattern, and the past-event hiding all look solid. A few things I'd like to address before approving:                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                          
1. Filter bar                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
Right now Offers / Needs / Events sit in the same pill row as All / New / Online / Recurrent / Weekend, separated only by a 1px divider for frontend (and also in the mobile). They're semantically different controls but read as identical chips, and all already exists as a behavioural reset which contradicts with "no type selected = all types." Could we lift the type chips into their own row or a segmented control above the filter bar (with a small "Type" / "Filter" label or just clear visual separation)? Would also be a good chance to make the row feel a bit more deliberate visually and improve UI/UX                                                             
                                                                                                                                                                                                                                                                                                                                                                          
2. cancel-event reason                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                          
The modal disables Confirm until cancelReason.trim().length > 0 (ServiceDetailPage.tsx:2329), but views.py:2324 reads request.data.get('reason', '') and services.py:1687 defaults to reason: str = ''. 

A mobile/curl client can cancel an event with no reason and write empty strings into every active participant's cancellation_reason. Either return 400 when  blank server-side (with a unit test).                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                         
3. displayServices                                                                                                                                                                                                                                                                                                                                              
                                         
DashboardPage.tsx:535-565 rebuilds two arrays and the comparator on every render, every keystroke in the search box re-filters and re-sorts the full feed. Wrap in useMemo keyed on [services, isAuthenticated, handshakeMap, activeTypes].

mzyavuz:
### 1. Filter Bar Redesign — `frontend/src/pages/DashboardPage.tsx`

**Problem:** The type chips (Offers / Needs / Events) and behavioral filter chips (All / New / Online / Recurrent / Weekend) were sitting in the same pill row, separated only by a 1px divider. They looked identical but meant completely different things. Having "All" in the same row also contradicted the logic where no type selected = all types shown.

**What I changed:**
- Removed the type chips and the divider from the shared behavioral filter container.
- **Desktop:** Added a dedicated second row below the search bar. It has a `FiLayers` icon as a visual anchor on the left, followed by Offers / Needs / Events in their own labeled pill group. The behavioral filters (All / New / Online / Recurrent / Weekend) stay in the top bar row, unchanged.
- **Mobile web (`base`):** Replaced the single mixed scrollable row with two separate scrollable rows — type chips on top, behavioral filters below — so the two concerns are clearly stacked.

### 2. Server-Side Cancel Reason Validation — `backend/api/views.py`

**Problem:** The event cancellation modal in the frontend correctly disables Confirm until a reason is typed. However, the backend endpoint (`cancel-event`) accepted an empty or missing `reason` field with no complaint. A curl or mobile client could bypass the frontend and write empty strings into every participant's `cancellation_reason` field.

**What I changed:**
- In `views.py`, right after reading `reason` from the request, we now call `.strip()` and return a `400 VALIDATION_ERROR` immediately if the result is blank.
- The service layer (`services.py`) is untouched — its `reason: str = ''` default remains valid for internal/admin callers.

### 3. Integration Tests — `backend/api/tests/integration/test_event_browse_api.py`

Added a new test class `TestCancelEventReasonValidation` with three cases:

| Test | Expected |
|---|---|
| POST with no `reason` field | `400` |
| POST with `reason: "   "` (whitespace only) | `400` |
| POST with `reason: "Venue unavailable"` | `200`, event status → `Cancelled` |

All three pass.

### 4. Performance Fix — `frontend/src/pages/DashboardPage.tsx`

**Problem:** Three derived values were plain `const` assignments recomputed on every render, including on every keystroke in the search box (which updates `searchQuery` state), even though none of them depend on the search query.

- `ownServiceHandshakes` — flattened array of all incoming handshakes
- `myServices` — the logged-in user's own listings, used for sidebar counters
- `displayServices` — the full filtered and sorted feed (auth-aware cancellation filter + type filter + expiry filter + handshake status sort)

**What I changed:** Wrapped all three in `useMemo` with tight dependency arrays:

```
ownServiceHandshakes → [incomingMap]
myServices           → [allActiveServices, user]
displayServices      → [services, isAuthenticated, handshakeMap, activeTypes]
```

Keystrokes now only update `searchQuery`, which is in none of those dependency arrays, so the feed recomputation is skipped entirely on every keystroke.

mzyavuz:
Latest Changes:

Desktop: The behavioral filters and the FiLayers icon button are back in the same pill row, separated by the same 1px divider as before. The icon button is grey when no type is selected; turns green with a count badge (e.g. 2) when types are active. Clicking it opens a small dropdown with Offers / Needs / Events each with its color-coded dot and highlight. Clicking outside closes it.

Mobile web: Same pill row as before — behavioral filters, divider, then the FiLayers type button. Tapping it toggles the same dropdown.

sgunes16 (approved):
The PR covers all things which are considerable issues with last fixes. Thank you for work.

PR #449, opened 2026-05-01
Feat/email verification gate clean

### LINKED ISSUE #359: Block unverified accounts from creating offers
## Context

An authenticated user whose email is still unverified can currently open `/post-offer` and successfully create an `Offer`.

This should be blocked. Email verification is already part of the auth flow, and the UI already exposes verification-related state (`user.is_verified`) plus resend/verify flows. However, offer creation is still effectively gated only by authentication/onboarding on the frontend, and the backend service creation endpoint does not appear to reject unverified users.

Because this is a business-rule/security-style restriction, backend validation must be the source of truth. Frontend UX should also prevent the action early and explain why the user cannot continue.

## Scope

- Prevent unverified users from creating `Offer` services.
- Keep verified users able to create offers as before.
- Add a frontend guard/message so unverified users are not allowed to complete the `/post-offer` flow silently.
- Add a backend enforcement so direct API calls to `POST /api/services/` also fail for unverified users.

## Tasks

- Add backend validation in the service creation flow for `type === "Offer"` when `request.user.is_verified !== true`.
- Return a clear error response and message for unverified offer creation attempts.
- Update the frontend offer creation flow to block or redirect unverified users before submit, with a clear explanation and a path to resend/complete verification.
- Ensure the current verification banner / resend flow remains compatible with the new restriction.
- Confirm the restriction does not unintentionally break `Need` or `Event` creation unless the product decision is to apply the same rule there as well.

## Acceptance Criteria

- A logged-in but unverified user cannot successfully create an offer from the UI.
- A logged-in but unverified user cannot create an offer by calling `POST /api/services/` directly.
- A verified user can still create an offer successfully.
- The UI explains that email verification is required before posting an offer.
- The user has an obvious next step such as resending the verification email or going to the verification flow.

## Tests

Backend tests should be added:

- Integration test: unverified authenticated user receives a rejected response when posting `type="Offer"` to `POST /api/services/`.
- Integration test: verified authenticated user can still create an offer successfully.
- Guard/regression test: non-offer service creation behavior stays unchanged unless intentionally expanded.

Frontend tests should be added:

- E2E test: unverified user is blocked on the `/post-offer` flow and sees the expected verification-required message.
- E2E test: verified user can still complete the normal offer creation flow.
- If a route-level/component-level guard is introduced, add a focused component/unit test for the guard state if that is already a nearby testing pattern.

## Notes

- Frontend-only blocking is not sufficient.
- Backend enforcement is required to protect against manual API requests and future UI regressions.

yaseminsirin:
Closes #359 
# Require verified email for posting services and joining listings

## Summary

Authenticated users whose email is **not verified** could still create **Offer / Need / Event** listings and could **request help**, **offer help on a Need**, or **RSVP to an Event** via the UI or direct API calls. This PR enforces the same business rule everywhere: **backend is the source of truth**, with **matching UX** on the web app (`user.is_verified`).

---

## Problem

- Listing creation (`POST /api/services/`) needed a consistent rule for **Need** and **Event**, not only Offer.
- **Join / request** flows (`express interest`, **join event**) did not require verification, so behavior could diverge from product intent and could be exercised without the UI.

---

## Solution

| Layer | Behavior |
|--------|----------|
| **API** | Return **403** with code **`EMAIL_NOT_VERIFIED`** when an unverified user tries to create any service type **or** join/request via handshake endpoints. |
| **Web** | **Route guards** on `/post-offer`, `/post-need`, `/post-event` using `RequireVerifiedEmail`. **Service detail**: `VerificationRequiredModal` before express-interest / join-event calls; API **403** handled as fallback. |

---

## Backend changes

- **`api/views.py`**
  - **`_require_verified_email`**, **`_indefinite_article`** — shared validation helper and copy.
  - **`ServiceViewSet.create`** — require verified email for **`Offer`**, **`Need`**, and **`Event`** before type-specific checks.
  - **`ExpressInterestView.post`** — `POST /api/services/<id>/interest/`
  - **`HandshakeViewSet.express_interest`** — router alias under `/api/handshakes/...`
  - **`HandshakeViewSet.join_event`** — `POST /api/handshakes/services/<id>/join-event/`
- **`api/exceptions.py`**
  - **`ErrorCodes.EMAIL_NOT_VERIFIED`**
- **`api/tests/helpers/factories.py`**
  - **`UserFactory`** defaults to **`is_verified=True`**; tests that need unverified users pass **`is_verified=False`**.

---

## Frontend changes

- **`App.tsx`** — wrap **`/post-need`** and **`/post-event`** with **`RequireVerifiedEmail`** (aligned with **`/post-offer`**).
- **`VerificationRequiredModal.tsx`** (new) — modal on service detail: message, **`authAPI.sendVerification()`** resend, dismiss, backdrop close.
- **`ServiceDetailPage.tsx`** — gate **`handleExpressInterest`** and **`handleJoinEvent`** when **`user?.is_verified === false`**; handle **`EMAIL_NOT_VERIFIED`** from API.

---

### Backend

- **`test_service_api.py`** — unverified cannot create Offer/Need/Event; verified can create all three.
- **`test_handshake_api.py`** — unverified blocked on express interest (Offer, Need, router alias) and join-event.
- **`test_wikidata.py`** — test user verified where Offers are POSTed.

### Frontend

- **Vitest** — `RequireVerifiedEmail.test.tsx`, `VerificationRequiredModal.test.tsx`
- **Playwright** — `email-verification-gate.spec.ts` (post routes + join/request modal; stubs `send-verification` where needed)
- **`tests/e2e/helpers/auth.ts`** — optional third arg on **`loginAs`** for `/users/me` overrides (e.g. `is_verified: false`).

---

## Manual QA

1. **Unverified** user: `/post-offer`, `/post-need`, `/post-event` → verification screen, no form fields.
2. **Unverified** on service detail → **Request / Offer to Help / Join Event** opens modal; resend path works (or shows error if mail backend fails).
3. **Verified** user → unchanged happy paths.
4. **API** (unverified JWT): `POST /api/services/`, `POST .../interest/`, `POST .../join-event/` → **403**, **`code: EMAIL_NOT_VERIFIED`**.

---

## Notes

- Organizer-only flows that do not use these applicant endpoints are unchanged by design unless they hit the same APIs.
- Existing email verification banner and `/verify-email-sent` remain compatible.

---

## Checklist

- [x] Backend enforcement (API cannot be bypassed from UI regressions alone)
- [x] Frontend guards + clear copy and resend CTA
- [x] Backend integration tests + frontend unit + E2E where patterns exist


sgunes16:
Thanks for pr. It works very well.

sgunes16 (approved):

Issue #459, opened 2026-05-03
Umbrella: Redesign Ranking Pipeline and Service/Event Scoring


yusufizzetmurat:
## The Problem
Our current hot_score system is trying to do too many things at once: liquidity, discovery, trust, and engagement. One number cannot handle all four. The current formula compromises on all of them. Old listings disappear, new providers cannot break in, and established services are being buried by aggressive time decay.

We are moving to a three-phase ranking pipeline. We previously only implemented the middle phase. This PR implements the full layered approach to ensure fairness and better results.

### Phase 1: Candidate Generation (Liquidity)
We are formalizing the RankingPipeline class to handle initial filtering.
* **Existing Logic:** Maintain status filters, PostGIS proximity, and type filters.
* **New Logic:** Exclude capacity-full listings (100% full) and stale events where the scheduled_time has passed.

### Phase 2: Scoring (Trust and Discovery)
We are replacing calculate_hot_score and its batch equivalent with a new formula:
**HotScore = Quality * Activity * CapacityMultiplier**

* **Quality:** Uses the Wilson Lower Bound for reputation. This prevents new accounts with 3/3 stars from outranking established providers with 95/100 stars. (Closes #317)
* **Activity:** Driven by HoursExchanged (total provisioned hours) instead of account age. (Closes #302)
* **Capacity Multiplier:** A 1.5x boost for listings at 75-99% capacity to drive completion. (Closes #304)
* **Events:** Events get a dedicated formula based on RSVP velocity (last 7 days) and Organiser-specific reputation. (Closes #303)

**Note:** We are removing time decay from the score itself. Content rotation is now handled in Phase 3.

### Phase 3: Probabilistic Re-rank (Fairness and Discovery)
To solve the discovery problem without breaking the Phase 2 math, we will implement an 80/20 Thompson Sampling split (Closes #316):
* **Exploitation (80%):** Top scores from Phase 2.
* **Exploration (20%):** A pool for cold-start providers (<5 handshakes), under-shown services (no activity in 14 days), and stale recurring listings. (Closes #305)

## Audit and Observability
* **ScoreAuditLog:** New model to track P, N, C, and HoursExchanged inputs for every recalculation. (Closes #308)
* **Error Handling:** Stop silent exception swallowing in _update_service_hot_score. All failures must be logged. (Closes #450)
* **Search Weighting:** Implement scored ordering (Title 1.0 > Tags 0.8 > Description 0.4). (Closes #306 and #324)
[...]

Issue #450, opened 2026-05-01
Silent error swallowing across backend, web, and mobile


yusufizzetmurat:
We swallow exceptions in too many places without logging. This is the bug class that hides every other bug — when something breaks in production, we have no signal. There is one root cause and one fix pattern; I'm filing it as one issue so we tackle it as a single observability pass.

**1a. Hot-score updates fail silently** — `backend/api/signals.py:51-59`
`_update_service_hot_score` wraps `calculate_hot_score()` and the `Service.objects.update(...)` write in a bare `except Exception: pass`. No logger, no metric. If scoring throws, bad data, divide-by-zero, DB hiccup — the service keeps its old score and we never find out. Ranking is one of the platform's load-bearing features and we have zero observability on it.

**1b. Pervasive `.catch(() => {})` on the web** — at least these spots:
- `frontend/src/pages/UserProfile.tsx:370, 373, 374, 378, 389`
- `frontend/src/pages/PublicProfile.tsx:261, 265`
- `frontend/src/pages/ServiceDetailPage.tsx:536, 696`
- `frontend/src/pages/ForumPage.tsx:324`
- `frontend/src/pages/ForumCreateTopic.tsx:63`

When a request fails, there's no log, no toast, no fallback UI, the user sees an empty section and we have no signal in production. This has almost certainly been masking real backend regressions.

**1c. Badge spinner hangs forever on error** — `frontend/src/pages/UserProfile.tsx:374` and `frontend/src/pages/PublicProfile.tsx:265`
`userAPI.getBadgeProgress(...)` is called without a `.finally(() => setBadgesLoading(false))`. Combined with the empty `.catch`, a failed request leaves `badgesLoading` true and the spinner spins forever. This is distinct from #167 (data correctness on success); this is the failure path.

**1d. Mobile SecureStore failure silently signs the user out** 
 `mobile-client/src/api/storage.ts:16-26`
`getStoredTokens` wraps `SecureStore.getItemAsync` in a try/catch that returns `null` on any error, with no log. SecureStore can be briefly unavailable on cold start (known on some Android OEMs and after OS upgrades). When that happens the app treats the user as signed out and forces a relogin, we have no way to investigate why.

### Suggested fix

Adopt one convention across the three layers and apply it in a single PR:

[...]

PR #460, opened 2026-05-03
feat: restructure ranking to a three-phase pipeline

### LINKED ISSUE #292: [NFR-12a, NFR-12b] We need performance baselines for the event feed and search
NFR-12a says feed renders in 2 seconds, NFR-12b says search returns in 1 second for 10,000 events. Neither is tested anywhere. We should at least establish a baseline so we know when a change makes things significantly slower.

Start with warning-mode: run a `pytest-benchmark` or `locust` scenario against a seeded dataset of 1,000+ events, capture p50/p95, and publish as a CI artifact. Don't add a hard gate. Document the dataset size, concurrency model, and environment spec so results are reproducible.

**Acceptance criteria:**
- Reproducible benchmark runs in CI against a documented dataset size and environment.
- p50/p95 metrics published as a CI artifact on each run.
- Baseline documented in the repo once established.

### LINKED ISSUE #302: [FR-17b] Fix the Service Hot Score formula: wrong variable and undecided direction
**What is broken:**
The denominator in `calculate_hot_score()` uses `hours_since_creation`, the number of hours since
the service was created. The spec says the denominator should be `HoursExchanged`: the total
`provisioned_hours` across all *completed* handshakes for that service. These are fundamentally
different signals and produce opposite ranking behaviour.

On top of that, the formula direction, divide or multiply,  has never been explicitly decided.
The SRS uses division (scores decay over time). The HiveMind design document uses multiplication
(established, active services grow). You cannot design the rest of the ranking pipeline without
settling this first.

**What needs to happen:**
1. Decide: multiply or divide? Document the rationale in a comment in `ranking.py`.
2. Replace `hours_since_creation` with `HoursExchanged` (sum of `provisioned_hours` on completed
   handshakes for the service, not the user).
3. Update `test_hot_score_time_decay` in `test_ranking.py` — the test name and assertion will change
   depending on which direction is chosen.

**Why this matters before anything else:**
Every other ranking feature (event formula, Thompson Sampling, decay) is calibrated against the
hot score range this formula produces. Building on the wrong formula wastes all downstream work.

**Acceptance criteria:**
- `calculate_hot_score()` uses `HoursExchanged`, not time elapsed.
- A code comment explains the multiply-vs-divide choice and why.
- Existing unit tests updated; new tests verify HoursExchanged drives the formula.

### LINKED ISSUE #303: [FR-17c, FR-17d] Implement a dedicated Event Hot Score formula
**What is broken:**
Events are currently ranked using the same formula as offers and needs. The reputation signals pulled
into that formula explicitly *exclude* Event-type handshakes (see `ranking.py` lines 36–63), so an
event's hot score is actually driven by the organiser's reputation from their non-event activities.
This means a newcomer who organises a great event gets a low ranking simply because they have no
prior offer/need history.

Formulate and implement a new event hot-score formula. Take account of the current user formula aswell and check the rankings vs offer/needs.

### LINKED ISSUE #304: [FR-17e] Decide and document the urgency multiplier scope
**What is ambiguous:**
The 1.5× urgency multiplier for listings at 75–99% capacity exists in the code but only activates
for `type='Event'`. The spec says "Listings" without qualification. The existing test
`test_non_event_service_no_boost` encodes the events-only behaviour as *intentional*.

In practice this is low-risk but the test and the spec are in direct conflict, and leaving it
unresolved means the test is either protecting intentional behaviour or guarding a bug.

**What needs to happen:**
Make a product decision and encode it clearly in both code and tests:
- **Events only:** add a comment in `ranking.py` explaining why non-events are excluded.
  Keep `test_non_event_service_no_boost` as-is.
- **All listing types:** update the code to check `current_participants / max_participants`,
  update `test_non_event_service_no_boost` to assert the boost *does* apply (rename it),
  and add a test for offers/needs with `max_participants > 1`.

**Acceptance criteria:**
- The scope decision is documented in `ranking.py` with a one-line comment.
- Tests match the agreed behaviour; no test is deleted without replacement.

### LINKED ISSUE #305: [FR-17f] Recurring listings without engagement growth should decay 10% per week
**What is missing:**
Nothing in the codebase touches `schedule_type` for ranking purposes. `ranking.py` has no decay
logic, and `update_hot_scores` applies no recurrence-based adjustment.

A `Recurrent` service whose engagement has not grown in the past 7 days should have its `hot_score`
multiplied by 0.9. A `One-Time` service is never subject to decay.

**What "lacking growth" means (needs further confirmation)**
A reasonable working definition: compare the count of completed/attended handshakes in the last 7
days against the count in the previous 7-day window. If the recent window is not larger, the service
is stale. This definition should be confirmed or replaced before implementation.

**Where the logic belongs:**
The `update_hot_scores` management command, it already runs on a schedule and has access to all
services. A `last_decay_applied_at` field on `Service` would prevent double-applying within one
7-day window.

**Acceptance criteria:**
- `Service` has a `last_decay_applied_at` DateTimeField (nullable).
- The management command decays stale recurring services by 0.9× at most once per 7-day window.
- One-time services are never decayed.
- A recurring service that does show growth is not decayed.
- The definition of "growth" is documented in code.

### LINKED ISSUE #306: [FR-17g, FR-SEA-01] Search results must be ordered by match quality, not just filtered
**What is missing:**
`search_filters.py` is a pure filter pipeline, it returns matching services but does not score or
rank them by how well they match. A service whose title exactly matches the query is returned in
the same unordered set as one whose description merely contains the keyword.

**What the spec requires:**
A three-tier weighted hierarchy:
- **1.0** — title contains the query term
- **0.8** — a WikiData tag associated with the service matches the query
- **lower** — description/body match only
- **0.5** — social proximity bonus (services from friends-of-friends via completed handshakes)

The current `TextStrategy` uses a single `Q(title__icontains=...) | Q(description__icontains=...) | Q(tags__name__icontains=...)` with equal implicit weight.

**Design decision required:**
Moving from flat filtering to scored ordering requires one of:
- Django's `SearchRank` + `SearchVector` (PostgreSQL full-text search, weight per field)
- An annotated numeric score per match tier, ordered by annotation
- BM25 or a dedicated search backend (Elasticsearch/Typesense)

The social proximity component requires the social graph and should be a no-op
placeholder with a `TODO` until that is built.

**Acceptance criteria:**
- Title matches appear before tag-only matches in search results.
- Tag matches appear before description-only matches.
- `TITLE_WEIGHT` and `TAG_WEIGHT` are defined as named constants in `search_filters.py`.
- Social proximity weight is wired in as `0` with a `TODO` comment.
- `TestSearchWeighting` in `test_search_filters.py` passes.

### LINKED ISSUE #307: [NFR-17a] Add a performance test that enforces the 1-second feed SLA
**What is missing:**
No test measures the time taken to return the discovery feed. An index on `(status, hot_score DESC)`
exists, but it has never been benchmarked against a realistic catalogue size.

**What needs to happen:**
A benchmark test (or a CI step) that seeds ~1 000 Active services, issues a hot-score-sorted feed
request, and asserts the response arrives in under 1 second. This test should run in the Docker
environment where PostGIS is available, not in the local unit test suite.

If the SLA is not met, the test output should point toward likely causes: missing composite index,
N+1 serializer calls, or missing `select_related`/`prefetch_related` in the view.

**Acceptance criteria:**
- `TestRankingQueryPerformance` in `test_ranking_api.py` passes consistently in Docker.
- The 1-second threshold is a named constant so it can be adjusted.

### LINKED ISSUE #308: [NFR-17c] Score inputs and recalculation timestamps must be auditable
**What is missing:**
When a service's `hot_score` changes there is no record of why. There is no `score_updated_at`
timestamp and no audit trail showing what values of P, N, C, and HoursExchanged were used.

**What needs to happen:**
At minimum:
- Add `score_updated_at = DateTimeField(null=True)` to `Service` and update it on every recalculation.
- Decide whether a full audit log (`ScoreAuditLog` model with inputs + timestamp) is needed or
  whether the timestamp alone satisfies the requirement.

A full audit log is more useful for debugging ranking anomalies but adds write overhead. A timestamp
alone is cheap but only tells you *when*, not *why*. This tradeoff should be discussed before
implementation.

**Acceptance criteria:**
- `Service.score_updated_at` is populated on every hot score recalculation.
- If a `ScoreAuditLog` is added: records include `positive_rep_count`, `negative_rep_count`,
  `comment_count`, `hours_exchanged`, and `recorded_at`.
- `TestScoreAuditability` in `test_ranking_api.py` passes.

### LINKED ISSUE #316: [FR-17i] Thompson Sampling for discovery feed fairness (80/20 exploitation/exploration)
**The problem:**
A purely deterministic hot-score feed creates a "rich get richer" loop: high-reputation providers
permanently dominate the feed and new providers never get organic visibility.

**What the design proposes:**
Route 80% of feed requests to high-score results (exploitation) and 20% to listings from users with
little or no transaction history (exploration). This is a Multi-Armed Bandit approach — Thompson
Sampling specifically.

The exploration bucket should draw from users whose total completed handshake count is below a
configurable threshold (e.g., < 5). When a request lands in the exploration bucket, the response
includes a mix of high-score results and at least one exploration candidate.

**Design questions to resolve before implementation:**
- Is the 80/20 split per-request (randomised on each API call) or sticky per user session?
- How is the exploration candidate selected — random draw, or highest-scoring among eligible users?
- Does the exploration slot appear at a fixed position (e.g., position 5) or scattered randomly?

**Where it sits:**
A post-scoring mixing layer in `ServiceViewSet.list()`, after the hot-score ordering is applied
but before pagination. The exploration threshold must be a Django setting, not a hardcoded constant.

**Acceptance criteria:**
- `should_explore(request)` returns True roughly 20% of the time (configurable).
- When in exploration mode, the response includes at least one listing from a user below the threshold.
- The exploration threshold and the 80/20 split are both configurable via Django settings.
- The per-request vs per-session decision is documented in code.

### LINKED ISSUE #317: [FR-17j] Wilson Score confidence interval for reputation fairness
**The problem:**
The hot score uses raw positive reputation counts. A user with 3 positives from their first and
only handshake has the same raw P=3 as a user with 300 positives from 350 handshakes. The ranking
formula treats them identically, allowing a lucky newcomer to mathematically outrank a high-volume
provider.

**What the design proposes:**
Replace raw counts with the Wilson Score lower-bound confidence interval. This estimates the true
positive rate with a confidence penalty for small samples:

```
lower_bound = (p_hat + z²/2n − z × sqrt((p_hat×(1−p_hat) + z²/4n) / n)) / (1 + z²/n)
where p_hat = positives / total, n = total, z = 1.96 (95% confidence)
```

A user with 3/3 scores around 0.29 (low confidence). A user with 300/350 scores around 0.80
(high confidence). New providers get a fair but honest ranking.

**Acceptance criteria:**
- `wilson_score_lower_bound(positives, total, z=1.96)` exists in `ranking.py`.
- Returns `0.0` when `total == 0` (no interactions yet), no exception raised.
- Hot score numerators use the Wilson-adjusted value instead of raw P.
- `TestWilsonScoreConfidenceInterval` in `test_ranking.py` passes.

### LINKED ISSUE #324: [FR-19f, FR-17g] Implement weighted search result ordering (title > tags > proximity)
**What is missing and duplicate:**
This requirement is identical to FR-17g. `search_filters.py` uses a flat `icontains` filter with
no scored ordering. A service whose title contains the query ranks identically to one whose
description merely contains it.

Opening it here as a tracking issue to link the two requirements.

### LINKED ISSUE #325: [NFR-19a] Add integration test enforcing the 2-second feed SLA
**What is missing:**
`@track_performance` in `performance.py` logs slow calls but does not fail a request or a test.
There is no test that seeds a realistic catalogue and asserts the feed responds within 2 seconds.

See FR-17/NFR-17a for the overlapping 1-second ranking SLA test. The discovery feed SLA (2s)
is a separate concern: it includes serialization, pagination, and permission checks on top of
the ranking query.

**Acceptance criteria:**
- A test seeds 1 000 Active services of mixed types, requests the discovery feed as an
  anonymous user, and asserts the response arrives in < 2 seconds.
- The threshold is a named constant.
- The test runs in Docker (requires PostGIS and Redis).

### LINKED ISSUE #371: Restrict ranking debug panel to admin dashboard with user simulation
## Problem

PR #364 introduced the ranking debug panel accessible to **all authenticated users** when the feature flag is enabled. The `debug-ranking` and `debug-ranking-availability` endpoints only require `IsAuthenticated`, which exposes internal ranking signals (reputation counts, social network structure, hot score formulas, social proximity boosts) to every user. This information could be used to game the ranking system.

### 1. Move debug panel to admin dashboard only
- Remove the floating debug launcher from the regular user dashboard (`DashboardPage.tsx`)
- The ranking debug panel should live exclusively inside the **Admin Dashboard** under a dedicated tab or section
- Regular users should never see the debug UI or be able to call the debug endpoints

### 2. Backend: restrict endpoints to admin roles
- `POST /api/services/debug-ranking/` should require admin role, not just authentication
- `GET /api/services/debug-ranking-availability/` should require admin role
- The `PlatformSetting` feature flag can remain as a master on/off, but the endpoints themselves must enforce admin access

### 3. Add "Simulate rankings as user" capability
- In the admin debug panel, add a user selector (search by name/email) that lets the admin pick a user to simulate
- When simulating, the backend computes ranking signals **from that user's perspective** (their social network, their location context, their search context)
- This is **read-only ranking simulation only** — it must NOT grant access to the simulated user's private data (messages, chat rooms, personal settings, etc.)
- The `build_service_debug_payload` should accept an explicit `viewer_user_id` parameter (admin-only) instead of always using `request.user`
- The admin can then compare how rankings differ across users to understand and debug the recommendation system

### 4. Admin UI flow
- Admin navigates to the ranking debug section in the admin dashboard
- Selects a user to simulate (or defaults to their own perspective)
- Sees the service feed as that user would see it, with the debug panel showing ranking breakdowns
- Can hover/select service cards to inspect individual ranking signals
- Sankey diagram and formula snapshot reflect the simulated user's perspective

## Additional cleanup
- Add debounce (~300ms) to the hover-triggered debug API calls to avoid excessive backend load
- Consider whether the `PlatformSetting` global toggle is still needed once access is admin-only, or if it should control something else (e.g., whether ranking signals are computed at all)

## Context
This was identified during review of #364. The current implementation is acceptable for demo purposes but should be hardened before any broader release.

yusufizzetmurat:
## Summary

  Restructures ranking from a single buggy hot_score formula into a three-phase pipeline (filter -> deterministic score -> probabilistic re-rank). One coherent answer instead of
  fourteen sequential half-answers.

  Phase 2 service formula: Quality x Activity x CapacityMultiplier where Quality is the Wilson lower bound on owner-by-type rep, Activity is log2(2 + lifetime HoursExchanged) +
  0.5 * log2(2 + comments), and CapacityMultiplier is 1.5 at 75-99% fill for events and group offers. No time decay -- customer rejected burying old listings.

  Phase 2 event formula: log2(2 + RSVPs_last_7d) x OrganiserQuality x CapacityMultiplier. Velocity-based, event-scoped quality so a great organiser doesn't get punished for never
  running an offer.

  Phase 3: Thompson Sampling. 80/20 exploit/explore. Exploration bucket includes cold-start providers, under-shown quality services (no completed handshake in 14d), and stale
  recurring services (flagged daily by check_recurring_growth). Slot fixed at position 5.

  Closes #292 
  Closes #302 
  Closes #303 
  Closes #304 
  Closes #305 
  Closes #306 
  Closes #307 
  Closes #308 
  Closes #316 
  Closes #317 
  Closes #324 
  Closes #325 
  Closes #371 
  Closes #450 
  Closes #459

  ## Test plan

  - [ ] make test (backend unit + integration; one pre-existing MinIO flake unrelated)
  - [ ] cd backend && .venv/bin/python -m pytest api/tests/integration/test_ranking_api.py
  - [ ] python manage.py update_hot_scores (backfill scores + audit rows)
  - [ ] python manage.py check_recurring_growth
  - [ ] manual: GET /api/services/?ordering=-hot_score returns sane scores
  - [ ] manual: POST /api/services/debug-ranking/ as non-admin returns 403; as admin returns 200


sgunes16:
Could you please create detailed issue for showing this implementation on recommendation debug bar. It would be impressive if we can show these formulations. You can assign me for that or you can implement it.

diclenaz7 (approved):
Don't see any problems

PR #466, opened 2026-05-04
fix(web,mobile): keep websockets reconnecting + render hygiene

### LINKED ISSUE #451: Real-time connections silently die for the rest of the session
Both the web and mobile real-time stacks have lifecycle bugs that leave the user disconnected with no UI signal. Once they trip, chat and live notifications stop arriving until full app restart.

### Sub-issues

**2a. Web `useWebSocket` reconnect counter never resets** — `frontend/src/hooks/useWebSocket.ts:93-115`
`onclose` increments `reconnectAttempts`, and once it reaches 5 the hook returns `prev` and stops scheduling reconnects. There is no codepath that resets the counter back to 0 when a connection successfully opens. So a tab that has had 5 lifetime drops — even spread over hours — eventually exhausts the budget and stays disconnected for the rest of the session, with no indicator.

**Repro:** Open the app, kill wifi for ~10 seconds, restore. Repeat 6 times. After the 6th drop, `setReconnectAttempts` returns `prev` and the socket never reconnects. Chat messages and live notifications stop arriving until full page reload.

**2b. Mobile notification socket stalls after backgrounding** — `mobile-client/src/hooks/useNotificationSocket.ts:93-110`
On foreground, `reconnectAttemptsRef.current` is reset to 0, but the code only kicks off a fresh `connect()` when the socket is already in `CLOSED`. iOS often keeps the socket in `CLOSING` while suspending the connection — in that state we reset the counter but never actually schedule a reconnect, so the user gets no real-time notifications until they restart the app.

### Suggested fix

- **2a:** reset `reconnectAttempts` to 0 inside the `onopen` handler once the connection has been stable for >2s. Consider showing a "Connection lost — retry" banner when we hit the cap so the user can act on it instead of silently degrading.
- **2b:** on foreground, if the socket is `CLOSING` or null, force-close it and call `connect()` after a short timeout. The reset+reconnect should be atomic.

### LINKED ISSUE #454: Frontend render hygiene, duplicate fetch and missing effect deps
Two small but real React-hygiene bugs in the same area of the codebase. Both look like merge artifacts that were never cleaned up.

a. `UserProfile` runs the same handshake fetch twice on every mount** — `frontend/src/pages/UserProfile.tsx:375-394`
Inside the same `useEffect`, `handshakeAPI.list(ac.signal)` is called and `setEventHandshakes(...)` is set on lines 375-379, and then the *exact same call and state update* happens again on lines 390-394. `setEventsLoading(true)` is also called twice.

**Repro:** Open DevTools → Network. Visit `/profile`. You'll see two `GET /api/handshakes/` requests fire back-to-back.

**b. `MapView` adds WebGL listeners on every render** — `frontend/src/components/MapView.tsx:429-441`
The `useEffect` that registers `webglcontextlost`/`webglcontextrestored` listeners closes with `})` instead of `}, [])`. So the effect re-runs on every render: cleanup → add → cleanup → add. Functionally fine because the cleanup matches, but it's wasted work and risks subtle ordering issues if a re-render lands while the GPU context is mid-transition.

### Suggested fix

- **a:** delete lines 390-394; the earlier block already handles the events tab.
- **b:** add an empty dependency array, or `[mapRef.current]` if we want it to re-bind when the ref changes.

yusufizzetmurat:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  Two small but real bug clusters that compound silently in production.                                                                                                                                              
                                                                              
  ### #451 — Real-time connections silently die                                                                                                                                                                      
                                                                              
  - **Web (`useWebSocket.ts`)** — the `reconnectAttempts` budget was reset eagerly inside `onopen`. A flapping connection (`open → close < 2s → open`) refilled the credits indefinitely, hiding the underlying      
  instability. The reset is now gated behind a 2-second stability timer that gets cleared on close or disconnect.
  - **Mobile (`useNotificationSocket.ts`)** — on foreground, the handler now treats any non-OPEN socket as broken: it force-closes the existing socket with code 1000 (so the original `onclose` skips               
  `scheduleReconnect`) and immediately calls `connect()`. This fixes the iOS quirk where the socket sticks in CLOSING after backgrounding and the user gets no notifications until app restart.                      
                                                                                                                                                                                                                     
  ### #454 — Frontend render hygiene                                                                                                                                                                                 
                                                                                                                                                                                                                     
  - `UserProfile.tsx` was running the same `handshakeAPI.list` call twice in the same effect. Removed the duplicate block.                                                                                           
  - `MapView.tsx`'s WebGL context-loss/restore listener effect was missing its dependency array (`})` instead of `}, [])`), so it cleanup-and-rebind every render. Added `[]`.                                       
                                                                                                                                                                                                                     
  Closes #451                                                                                                                                                                                                        
  Closes #454                                                                                                                                                                                                        
                                                                                                                                                                                                                     
  ## Test plan                                                                
                                      
  - [ ] **Repro #451 (2a)**: open the app, kill wifi for ~10s, restore — repeat 6×. After the 6th drop, chat/notifications still arrive.                                                                             
  - [ ] **Repro #451 (2b)**: background the iOS app, leave for >30s, foreground — push notifications continue to arrive without an app restart.
  - [ ] **Repro #454 (a)**: open DevTools → Network on `/profile`. Only one `GET /api/handshakes/` request fires per mount.                                                                                          
  - [ ] **Repro #454 (b)**: profile MapView mounts. Listener bind/unbind events fire once, not on every render.                                                                                                      
  - [ ] `cd frontend && npm run build` clean.                                                                                                                                                                        
  - [ ] `cd mobile-client && npm test` green.

### DIFF
--- a/frontend/src/components/MapView.tsx
+++ b/frontend/src/components/MapView.tsx
@@ -438,7 +438,7 @@ export function MapView({ services, height = '400px', onServiceClick, userLocati
       canvas.removeEventListener('webglcontextlost', onLost)
       canvas.removeEventListener('webglcontextrestored', onRestored)
     }
-  })
+  }, [])
 
   // Fly to user location only on the FIRST time it becomes available
   const didFlyToUser = useRef(false)
--- a/frontend/src/hooks/useWebSocket.ts
+++ b/frontend/src/hooks/useWebSocket.ts
@@ -32,6 +32,7 @@ export function useWebSocket({
   const [reconnectAttempts, setReconnectAttempts] = useState(0)
   const wsRef = useRef<WebSocket | null>(null)
   const reconnectTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+  const stableTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const openedAtRef = useRef<number>(0)
   const connectedUrlRef = useRef<string | null>(null)
   const onMessageRef = useRef(onMessage)
@@ -69,7 +70,14 @@ export function useWebSocket({
       ws.onopen = () => {
         openedAtRef.current = Date.now()
         setIsConnected(true)
-        setReconnectAttempts(0)
+        // Reset the reconnect budget only after the connection has been stable
+        // for >2s, so flapping connections (open→close→open) don't keep
+        // refilling credits and masking a chronically broken socket.
+        if (stableTimerRef.current) clearTimeout(stableTimerRef.current)
+        stableTimerRef.current = setTimeout(() => {
+          setReconnectAttempts(0)
+          stableTimerRef.current = null
+        }, 2000)
         onOpenRef.current?.()
       }
 
@@ -93,6 +101,10 @@ export function useWebSocket({
       ws.onclose = (event) => {
         setIsConnected(false)
         connectedUrlRef.current = null
+        if (stableTimerRef.current) {
+          clearTimeout(stableTimerRef.current)
+          stableTimerRef.current = null
+        }
         onCloseRef.current?.()
 
         // 4xxx = application-level rejection — never retry.
@@ -122,6 +134,10 @@ export function useWebSocket({
     if (reconnectTimeoutRef.current) {
       clearTimeout(reconnectTimeoutRef.current)
       reconnectTimeoutRef.current = null
+    }
+    if (stableTimerRef.current) {
+      clearTimeout(stableTimerRef.current)
+      stableTimerRef.current = null
     }
     if (wsRef.current) {
       wsRef.current.close(1000)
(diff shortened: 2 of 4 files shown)


mzyavuz (approved):
Test coverage
The test plan items you verified manually:

#451 (2a) — wifi drop × 6: First message arrives after a few seconds delay — pass (reconnect works, just takes a moment to re-establish).
#451 (2b) — iOS background: Not testable locally, acknowledged.
#454 (a) — single handshake request: Two requests seen in dev = Strict Mode, not a bug. Pass.
#454 (b) — MapView listener once: Requires console instrumentation to verify formally.

Approve. The logic is correct, the fixes are minimal, and all three clear-cut regressions (#451 flapping, #451 iOS CLOSING, #454 double fetch, #454 listener rebind) are addressed properly. The double request in dev is a Strict Mode artifact, not a regression from this PR. (Nevertheless, at the prod I have seen that there are 2 double request (total 4 request), it has been suggested that double request is not bug so the problem will be solved after deployment I think)

PR #468, opened 2026-05-04
fix(timebank): defer cache invalidation to transaction.on_commit

### LINKED ISSUE #348: Cache invalidation holds DB lock in `provision_timebank` and `cancel_timebank_transfer`
### Summary

  `provision_timebank` and `cancel_timebank_transfer` in `backend/api/utils.py` call Redis cache
  invalidation functions **inside** `transaction.atomic()`, while the `select_for_update()` row
  lock on the user is still held. `complete_timebank_transfer` already does this correctly via
  `transaction.on_commit()` the other two should match.

  ### Problem

  ```python
  # provision_timebank — lines 87–123
  with transaction.atomic():
      receiver = User.objects.select_for_update().get(...)  # row lock acquired
      ...
      invalidate_conversations(str(receiver.id))   # Redis network call — lock still held
      invalidate_transactions(str(receiver.id))    # Redis network call — lock still held
  ```

  Two concrete issues:

  1. Extended lock duration: invalidate_transactions runs a Redis SCAN + DEL across
  all matching keys. That network roundtrip happens while the PostgreSQL row lock on the user
  is held, blocking any concurrent request that needs to modify the same user's balance.
  2. Cache invalidated on rollback: if the transaction rolls back after the Redis call, the
  cache is already cleared. Functionally harmless but semantically wrong; the invalidation
  should only happen if the commit succeeds.

  **Expected behaviour**

  Cache invalidation should be deferred to transaction.on_commit(), as already done in
  complete_timebank_transfer (line 179):

  transaction.on_commit(invalidate_after_commit)

  **Affected functions**
provision_timebank      
cancel_timebank_transfer

  **Fix**

  Move the invalidate_* calls in both functions into a transaction.on_commit() callback,
  matching the pattern in complete_timebank_transfer.

  **Out of scope**

  - No behaviour changes, only when the Redis calls are scheduled
  - No changes to complete_timebank_transfer, which is already correct

yusufizzetmurat:
## Summary                                                          
                                                                                                                                                                                                                     
  `provision_timebank` and `cancel_timebank_transfer` were calling Redis `invalidate_conversations` / `invalidate_transactions` *inside* `transaction.atomic()` while the `select_for_update()` row lock on the user 
  was still held. That had two real consequences:                                                                                                                                                                    
                                                                                                                                                                                                                     
  1. Stretched lock window — `invalidate_transactions` runs a Redis `SCAN + DEL` over matching keys; that network round-trip was happening with the Postgres row lock open, blocking concurrent writes against the
  same user.                                                                                                                                                                                                         
  2. Cache invalidated on rollback — if the surrounding transaction rolled back, we'd already have purged Redis for state that never persisted.
                                                                                                                                                                                                                     
  `complete_timebank_transfer` already does this correctly via `transaction.on_commit()`. This PR brings the other two functions onto the same pattern.
                                                                                                                                                                                                                     
  Closes #348                                                                                                                                                                                                        
                                                      
  ## Files                                                                                                                                                                                                           
                                                                                                                                                                                                                     
  - `backend/api/utils.py` — wrap the Redis calls in `transaction.on_commit()` in both functions                                                                                                                     
  - `backend/api/tests/integration/test_timebank_cache_invalidation.py` — 4 new tests:                                                                                                                               
    - provision: invalidations run after commit, not during the lock                                                                                                                                                 
    - provision: invalidations skipped on rollback                             
    - cancel: invalidations run after commit                          
    - cancel: invalidations skipped on rollback                       
                                                                                                                                                                                                                     
  ## Test plan                                        
                                                                                                                                                                                                                     
  - [ ] `cd backend && .venv/bin/python -m pytest api/tests/integration/test_timebank_cache_invalidation.py -v` — 4 green                                                                                            
  - [ ] `cd backend && .venv/bin/python -m pytest api/tests/unit/test_utils.py api/tests/integration/test_handshake_api.py -v` — existing timebank tests still green (no regressions)                                
  - [ ] `make test` — full backend suite green; coverage ≥ 70

### DIFF
--- a/backend/api/tests/integration/test_timebank_cache_invalidation.py
+++ b/backend/api/tests/integration/test_timebank_cache_invalidation.py
@@ -1,0 +1,98 @@
+"""Integration tests for #348 — Redis invalidation in provision_timebank
+and cancel_timebank_transfer must run on commit, not while the row lock is
+held, and must NOT run when the surrounding transaction rolls back.
+"""
+from decimal import Decimal
+from unittest.mock import patch
+
+import pytest
+from django.db import transaction
+
+from api.tests.helpers.factories import HandshakeFactory, ServiceFactory, UserFactory
+from api.utils import cancel_timebank_transfer, provision_timebank
+
+
+def _build_offer_handshake(receiver_balance: Decimal = Decimal('5.00')):
+    provider = UserFactory(timebank_balance=Decimal('5.00'))
+    receiver = UserFactory(timebank_balance=receiver_balance)
+    service = ServiceFactory(user=provider, type='Offer', duration=Decimal('2.00'))
+    handshake = HandshakeFactory(
+        service=service,
+        requester=receiver,
+        status='accepted',
+        provisioned_hours=Decimal('2.00'),
+    )
+    return provider, receiver, service, handshake
+
+
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.integration
+class TestProvisionTimebankInvalidation:
+    """provision_timebank wraps Redis invalidation in transaction.on_commit()."""
+
+    def test_invalidations_run_after_commit_on_success(self):
+        _, _, _, handshake = _build_offer_handshake()
+
+        with patch('api.utils.invalidate_conversations') as inv_conv, \
+             patch('api.utils.invalidate_transactions') as inv_tx:
+            with transaction.atomic():
+                provision_timebank(handshake)
+                # While still inside the atomic block, neither callable
+                # should have run yet — they're queued via on_commit.
+                inv_conv.assert_not_called()
+                inv_tx.assert_not_called()
+
+            # After the outer transaction commits, on_commit fires.
+            assert inv_conv.call_count == 2  # receiver + provider
+            assert inv_tx.call_count == 1    # receiver only on provision
+
+    def test_invalidations_skipped_on_rollback(self):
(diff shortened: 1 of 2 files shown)


mzyavuz (commented):
Hey Yusuf, nice fix! The root cause is clearly explained in the PR description and the change itself is minimal and correct. A few notes:

**`utils.py`**
- Capturing `receiver_id` / `provider_id` as strings before the closure is the right call — good instinct to avoid lazy ORM access inside `on_commit`.
- The comments are useful since the "why" isn't obvious, but they're a bit long. One line each would do.

**Tests**
- `@pytest.mark.django_db(transaction=True)` is essential here — without it `on_commit` never fires in tests, so good catch on that.
- The 4-test matrix (provision/cancel × commit/rollback) covers exactly what matters.
- One thing: `assert inv_tx.call_count == 1` in the provision test — could use a short comment explaining why only the receiver gets `invalidate_transactions` and not the provider. Without it, a future reader might think it's a missing call.

**One minor thing to be aware of:**
If `provision_timebank` or `cancel_timebank_transfer` ever get called outside an atomic block, `on_commit` fires immediately (Django's documented behavior). That's fine and probably won't happen given how the views are structured, but it's untested.

Overall looks good to me, happy to approve once you consider the comment on the `call_count == 1` assertion. 

**Approved ✓**

mzyavuz (approved):

Issue #269, opened 2026-03-28
Feature 11 (NFR-11a, NFR-11c): add performance and usability test coverage for event flows


yusufizzetmurat:
**Problem**: No automated thresholds for event create/update latency and no focused responsive/a11y checks.
**Decision required**:
  - Enforcement model:
    - Option A: hard CI gate on strict p95 thresholds
    - Option B: warning-only metrics trend first, hard gate after baseline
   
**Implementation spec**:
  - Define perf measurement protocol (warmups, dataset size, percentile, environment profile).
  - Add API timing probes for create/update operations and threshold assertions.
  - Add frontend responsive + accessibility checks (desktop/mobile breakpoints, keyboard flow, aria labels).
  - Add report artifact output in CI for trend review.
 
**Acceptance criteria**:
  - Documented and reproducible NFR test methodology.
  - Automated checks for event create/update latency and critical accessibility paths.
  - Baseline metrics captured and published in CI artifacts.
  - Migration plan from warning mode to gating mode.

yusufizzetmurat:
The performance and a11y instrumentation this issue asked for is now wired in PR #519:

- NFR-11a (event create p95 < 1.5 s) — `frontend/tests/perf/event-create.js`, run by `make test-perf` and the new `ci-e2e-nightly.yml` workflow.
- NFR-11c (a11y for event flows) — covered by `frontend/tests/e2e/a11y/baseline.spec.ts` via `expectNoBlockingA11y` from `helpers/axe.ts`. The baseline includes the event create modal alongside auth, dashboard, profile, forum, search and registration.

The CI job is on the cron `0 2 * * *` schedule and has not fired yet. Once one nightly run is green I'll attach the k6 summary and the axe report from the run artifacts here. Leaving the issue open until that artifact set lands.

yusufizzetmurat:
Folded into epic #579 under "Testing & Validation". Closing as duplicate; the performance + usability NFR coverage for event flows is the natural fit for the epic's testing track. The `6ef0416` test-infra foundations on dev are the starting point.

Issue #268, opened 2026-03-28
E2E (FR-11a..FR-11o, NFR-11a..NFR-11c): add requirement-mapped feature-11 suite


yusufizzetmurat:
**Problem**: Current event E2E tests are limited and not organized as one-spec-per-requirement.

**Implementation spec**:
  - Add `frontend/tests/e2e/feature-11/` with filename and title mapping by requirement ID.
  - Build shared event helpers (`createEvent`, `joinEvent`, `checkInEvent`, `cancelEvent`, role switching).
  - Mark known backend-gapped FRs (`FR-11g/h`) as expected pending or placeholder with skip reason.
  - Add matrix file mapping each E2E spec to requirement and proof assertion.
 
**Acceptance criteria**:
  - Spec names and titles include FR/NFR IDs.
  - Setup deterministic and isolated (no cross-test coupling).
  - Assertions are minimum proof and requirement-centric.
  - CI executes `feature-11` suite independently.

PR #470, opened 2026-05-04
feat(api): enhance service edit-locking and event filtering

### LINKED ISSUE #267: Feature 11 (FR-11f, FR-11n): unify event lockdown window semantics between API and UI
**Problem**: Backend edit-lock and frontend window utility diverge in post-start behavior.
**Decision required**:
  - Lock semantics after start:
    - Option A: lock remains active after start until terminal status
    - Option B: lock only in `[start-24h, start)` window

**Implementation spec**:
  - Define one canonical rule in backend and expose lock state in service payload (`edit_locked`, `edit_lock_reason`).
  - Frontend consumes backend lock flags instead of duplicating date math.
  - Add explicit UX messaging for locked state and admin exception behavior.
 
 **Acceptance criteria**:
  - UI edit affordance always matches API authorization.
  - No path where UI allows edit but API returns lock denial unexpectedly.
  - Regression coverage for `T-25h`, `T-1h`, `T=start`, `T+1h`.

### LINKED ISSUE #285: [FR-12c] Users should be able to filter events by date
Right now there's no way to filter events by when they're happening. If I'm looking for something to do this weekend or want to find events in a specific window, I have to scroll through everything and check dates manually. This makes the event feed pretty hard to use in practice.

We need to add `date_from` and `date_to` query parameters to the event listing. Both should be optional ISO-8601 dates. When provided, only events with a `scheduled_time` within the range should be returned. Non-event services don't have a meaningful `scheduled_time` so the filter should only kick in when `type=Event` is also set, or be silently ignored otherwise.

On the backend, the cleanest way to do this is to add a `DateRangeStrategy` to `search_filters.py` following the same Strategy Pattern already used for location, text, and type filtering. It should slot into `SearchEngine.__init__` after `TypeStrategy` so the queryset is already scoped to events before date filtering is applied. The `ServiceViewSet` should extract the new params and pass them into `SearchEngine.search()`.

On the frontend, we need actual UI controls for this. Predefined preset pills (`Today`, `This Week`, `This Weekend`, `Next Week`) would cover most real use cases and are much simpler than a full date picker. When a user picks a preset, the corresponding `date_from`/`date_to` values should be computed client-side and sent as query params.

The cache key for the service list already includes query params, so caching should continue to work correctly. Location queries that bypass Redis are unaffected.

**Acceptance criteria:**
- `GET /api/services/?type=Event&date_from=2026-04-01&date_to=2026-04-07` returns only events whose `scheduled_time` falls in that range.
- Events without a `scheduled_time` are excluded when the filter is active (they have no meaningful date).
- Invalid date strings return a `400` with a clear error rather than silently ignoring the param.
- Frontend date preset pills produce the correct `date_from`/`date_to` values and the filtered results update accordingly.
- Unit tests for `DateRangeStrategy`; integration tests for the API filter; E2E for the preset pill → filtered results flow.

### LINKED ISSUE #288: [FR-12g] Cancelled event detail page should say the event was cancelled
If you have a direct link to an event that got cancelled, a notification, a bookmark, or a shared URL, the page currently renders normally with all the join buttons visible. That's confusing.

When `service.status === 'Cancelled'`, the detail page should show a clear cancellation banner above the fold and suppress all join/evaluate/check-in actions. If the cancellation-reason field from issue FR-11l ever lands, it should be displayed here too.

**Acceptance criteria:**
- Navigating directly to a cancelled event's URL shows a clear cancellation notice above the fold.
- All join, check-in, and evaluate actions are hidden or disabled for cancelled events.
- The API still returns the object with `status: "Cancelled"` on a direct GET (so the frontend can render the notice rather than hitting a 404).
- Integration test: direct GET of a cancelled event returns 200 with `status=Cancelled`.
- E2E test: user navigating to a cancelled event URL sees the cancellation state, not a functional event page.

### LINKED ISSUE #321: [FR-19a] Add a dedicated events discovery view with category, date, and location filters
gap: **there is no dedicated event discovery page**.
Events are only surfable via the main service feed with a manual `?type=Event` filter that has
no first-class UI entry point. Users have no purpose-built place to browse, filter, and join events.

**What is missing:**
A `/events` page (or equivalent route) that:
- Shows only `type=Event` listings by default — no offers or needs mixed in.
- Exposes filters meaningful to events: date range (FR-12c), location/distance, and category/tag.
- Displays event-specific information on cards: scheduled date/time, participant count vs. capacity,
  and organiser name.
- Is linked from the main navigation so users can reach it in one click.

The backend already supports all of this (`GET /api/services/?type=Event&date_from=...&distance=...`).
This is a frontend-only gap.

**What needs to happen:**
- A new `EventsPage.tsx` (or `EventDiscoveryPage.tsx`) at route `/events`.
- Event cards reuse the existing card component but display event-specific fields.
- Filter bar includes: date range picker, distance slider, tag/category selector.
- The main navigation bar links to `/events`.
- The homepage hero or "How it Works" section links to `/events` as a call-to-action.

**Acceptance criteria:**
- `/events` route exists and shows only Event-type listings.
- Date range, distance, and tag filters work and update results without full reload.
- Empty state is shown when no events match the current filters.
- Navigation bar includes an "Events" link.
- Page is accessible to anonymous users (read-only, join requires login).

yusufizzetmurat:
## Summary                                                                                                                                                                                                         
                                                                               
  Bundles four event-feature gaps that share scope (search filters, ServiceSerializer, ServiceDetailPage, navigation).                                                                                               
                                                                                                                                                                                                                     
  ### #285 — Filter events by date (FR-12c)                                                                                                                                                                          
                                                                                                                                                                                                                     
  - New `DateRangeStrategy` in `api/search_filters.py`, slotted into `SearchEngine` after `TypeStrategy`. Honours `date_from` / `date_to` query params (ISO-8601 dates or full datetimes), only when `type=Event`.   
  - Bare-date `date_to` is normalised to end-of-day so users picking "today" actually see today's events.                                                                                                            
  - New `InvalidSearchParam` exception turns malformed dates into a 400 with a field-level error instead of silently ignoring them.                                                                                  
  - `ServiceViewSet.get_queryset` plumbs the new params through and converts `InvalidSearchParam` to a DRF `ValidationError`.                                                                                        
  - `TypeStrategy` now also accepts `'Event'` (was Offer/Need only).                                                                                                                                                 
  - Frontend `serviceAPI.list` accepts `date_from` / `date_to`.                                                                                                                                                      
                                                                                                                                                                                                                     
  ### #267 — Unify event lockdown semantics (FR-11f, FR-11n)                                                                                                                                                         
                                                                                                                                                                                                                     
  Backend and frontend were duplicating the date math with subtly different rules — backend treated events as locked from `start - 24h` onwards, frontend only inside `[start - 24h, start)`.                        
                                                                               
  - New `Service.edit_lock_reason` and `Service.edit_locked` properties on the model. Canonical rule: locked when `status` is terminal (`Completed`/`Cancelled`) **or** the event is within 24h of `scheduled_time`  
  **or** it has already started. Single source of truth.                       
  - `ServiceSerializer` exposes `edit_locked` / `edit_lock_reason` as read-only fields.                                                                                                                              
  - `ServiceDetailPage.tsx` consumes `service.edit_locked` first, falling back to the legacy `isWithinLockdownWindow(...)` only when the API hasn't shipped the new field yet.                                       
  - Frontend `Service` type gains `edit_locked` / `edit_lock_reason`.                                                                                                                                                
                                                                                                                                                                                                                     
  ### #288 — Cancelled event detail page says cancelled (FR-12g)                                                                                                                                                     
                                                                                                                                                                                                                     
  - New "above the fold" cancellation banner in `ServiceDetailPage.tsx` that fires for any `service.status === 'Cancelled'`. Existing inline cancellation messages remain as detail-area context.                    
  - API already returns the cancelled service on direct GET (verified by `test_cancelled_event_detail_returns_200`).
                                                                                                                                                                                                                     
  ### #321 — Dedicated events discovery view (FR-19a)                                                                                                                                                                
                                                                                                                                                                                                                     
  - New `frontend/src/pages/EventsPage.tsx` at `/events`. Defaults to `type=Event`. Date preset pills (Today / This Week / This Weekend / Next Week), distance presets (5/25/100 km — disabled until geolocation is  
  granted), event cards showing scheduled date, location area, capacity, and organiser.
  - Linked from the navbar (desktop + mobile) between Browse and Forum.                                                                                                                                              
                                                                                                                                                                                                                     
  ### #268 / #269 — Test coverage     
                                                                                                                                                                                                                     
  - Backend integration tests added for the above (`backend/api/tests/integration/test_events_feature_api.py`, 9 collected).                                                                                         
  - The full requirement-mapped E2E feature-11 suite (#268) and perf/usability NFR scaffolding (#269) are deferred to a `PR-C-tests` follow-up — they need fixture/auth helpers and CI baselining that's better
  reviewed independently.                                                                                                                                                                                            
                                                                               
  Closes #267                                                                                                                                                                                                        
  Closes #285                                                                  
  Closes #288                         
  Closes #321
  Refs #268
  Refs #269

  ## Files                                                                                                                                                                                                           
   
  **Backend**                                                                                                                                                                                                        
  - `api/models.py` — `Service.edit_lock_reason` + `edit_locked` properties    
  - `api/search_filters.py` — `DateRangeStrategy`, `InvalidSearchParam`, `'Event'` accepted in `TypeStrategy`                                                                                                        
  - `api/serializers.py` — `edit_locked` / `edit_lock_reason` on `ServiceSerializer`                                                                                                                                 
  - `api/views.py` — `ServiceViewSet` plumbs `date_from`/`date_to`, converts `InvalidSearchParam` → 400                                                                                                              
  - `api/tests/integration/test_events_feature_api.py` — 9 new tests                                                                                                                                                 
                                                                                                                                                                                                                     
  **Frontend**                                                                                                                                                                                                       
  - `src/services/serviceAPI.ts` — `date_from` / `date_to` params              
  - `src/types/index.ts` — `edit_locked`, `edit_lock_reason` on `Service`                                                                                                                                            
  - `src/pages/ServiceDetailPage.tsx` — cancellation banner + `edit_locked` consumption                                                                                                                              
  - `src/pages/EventsPage.tsx` — new page                                                                                                                                                                            
  - `src/App.tsx` — `/events` route                                                                                                                                                                                  
  - `src/components/Navbar.tsx` — desktop + mobile nav links                                                                                                                                                         
                                                                                                                                                                                                                     
  ## Test plan                        
                                                                                                                                                                                                                     
  - [ ] `cd backend && .venv/bin/python -m pytest api/tests/integration/test_events_feature_api.py -v` — 9 green                                                                                                     
  - [ ] `make test` — full backend suite green (no regressions in existing event tests)
  - [x] `cd frontend && npm run build` — clean                                                                                                                                                                       
  - [x] Manual: `GET /api/services/?type=Event&date_from=2026-04-01&date_to=2026-04-07` returns only events in range                                                                                                 
  - [x] Manual: `GET /api/services/?type=Event&date_from=junk` → 400 `{"date_from": "Invalid date format..."}`                                                                                                       
  - [ ] Manual: cancelled event detail page shows red banner above the fold; join/check-in/evaluate hidden or disabled                                                                                               
  - [ ] Manual: open `/events` from navbar — date pills filter the list, distance pills disabled until you allow location                                                                                            
  - [ ] Manual: as organiser, open an event whose start is < 24h away — service detail shows the lockdown reason from the API rather than the old fixed string


mzyavuz (changes requested):
## Review: Request Changes

Thanks for the bundle, the architecture and backend logic are solid, and the EventsPage is a good addition. A few things need to be fixed before this can merge:

### 1. Failing test (`test_invalid_date_returns_400`)

The test asserts `'date_from' in response.data`, but the actual response wraps field errors under a `field_errors` key:

```python
{'detail': 'Validation failed.', 'code': 'VALIDATION_ERROR', 'field_errors': {'date_from': [...]}}
```

The assertion needs to check `response.data['field_errors']` instead, or the test needs to be updated to match the actual error envelope this project uses. Either fix the test or fix the view to return a flat structure — but the two must agree.

### 2. Spots label is confusing

`17 / 20 left` reads as "17 out of 20 remaining", but a reader's first instinct is "17 of 20 total filled". The label should make it unambiguous. Suggestions:

- `3 spots left` (just the remaining count)
- `17 / 20 joined` (taken / capacity)

Pick one and keep it consistent with how capacity is shown elsewhere in the app.

### 3. EventsPage (date filter pills are not actually filtering)

The preset pills change UI state but the events list is not being filtered by date. The `date_from` / `date_to` params need to be verified end-to-end: check that `rangeForPreset` is producing the right values, that they are being passed to `serviceAPI.list`, and that the backend is receiving and applying them. Needs a fix + manual verification before this can merge.

### 4. Distance filter (user has no idea why it's disabled)

When geolocation hasn't been granted, the distance pills are greyed out silently. A user landing on this page for the first time will just see broken-looking buttons. There should be a visible prompt, even a single line like *"Allow location access to filter by distance"*, so they understand what to do.

### 5. Cancelled event (action buttons should be disabled) 

On a cancelled event's detail page, the cancellation banner correctly appears above the fold. However the **Complete** and **Cancel event** buttons remain enabled. Those actions are meaningless (and potentially error-prone) on a cancelled event and should be disabled or hidden when `service.status === 'Cancelled'`. The `edit_locked` flag is already available in the payload, use it to drive the button state.

### 6. EventsPage (color palette is too flat)

The page works but feels colourless compared to the rest of the app. Event cards could benefit from:

- A colour-coded date badge or left border accent to draw the eye
- A subtle capacity indicator (e.g. amber when nearly full, as the hot-score urgency multiplier already defines a 75% threshold)
- The organiser avatar or initials instead of plain text

This doesn't need to be a full redesign, even small accent colors would make it feel like a real discovery surface rather than a plain list.

Please address items 1-4 before re-requesting review. Item 5 and 6 is a strong suggestion but I'll leave the implementation to your judgement.

And of course, the conflicts should be resolved.

yusufizzetmurat:
Thank you for the review, I have decided to remove the EventsPage entirely since on the later PR activity page captures the idea more correctly. Other issues are fixed.

mzyavuz (changes requested):
Thanks for addressing the feedback. Quick status check on each point:

**Items resolved:**

- **1 (test assertion)** — `test_invalid_date_returns_400` now correctly checks `response.data.get('field_errors', {})`. This matches the envelope.
- **2, 3 & 4 (EventsPage date/distance filter)** removing `EventsPage.tsx` entirely is a valid call if the activity page supersedes it. The `date_from` / `date_to` plumbing in `serviceAPI.ts` and the backend can still be used by that future page, so nothing is wasted.
- **5 (cancelled event action buttons)** — the new conditional in the organiser-actions section (`service.status === 'Cancelled' ?` branch) replaces the Complete/Cancel buttons with an inline banner rather than leaving them enabled. That's the right approach.

**Needs to resolved:**
- **Merge conflicts**: The original review called these out and the PR still shows conflicts. These need to be resolved before this can go green.

Everything else looks clean. Once the conflicts are resolved and the spots-label question is confirmed, this is good to merge from my side.

Can you look at the #496, the notifications of the events has not worked as expected.

yusufizzetmurat:
The conflicts are resolved.

mzyavuz (approved):
All tests at the acceptance have passed. Conflicts are resolved. Approved.

PR #473, opened 2026-05-04
feat(mobile): offline cache for active handshakes and joined events 

### LINKED ISSUE #322: [FR-19b, NFR-19b] Mobile client: offline cache for active handshakes and joined events
**What is missing:**
The mobile client (`mobile-client/src/`) uses `expo-secure-store` for token persistence only.
There is no data cache for handshakes or events. If the device goes offline, the user sees
empty screens or errors rather than their active commitments.

The spec requires both offline read access **and** graceful resync when connectivity returns.
Neither is implemented.

**What needs to happen:**
- A local cache layer (e.g. MMKV, AsyncStorage, or SQLite via expo-sqlite) storing:
  - Active handshakes the user is party to
  - Events the user has joined
- A resync hook or background task that flushes stale data and re-fetches when the network
  becomes available (NetInfo-based).
- The cache should be invalidated, not silently stale: if cached data is older than a configurable
  TTL, the UI should indicate "last synced N minutes ago."

**Design decisions to make:**
- Which persistence library? 
- What is the cache TTL before the UI shows a stale indicator?
- Does resync happen in the foreground or as a background task?

**Acceptance criteria:**
- User can read their active handshakes and joined events while offline.
- Reconnection triggers a resync without user action.
- Stale data is visually indicated if cache exceeds TTL.
- Auth token is not conflated with data cache — they remain separate stores.

yusufizzetmurat:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  Closes the FR-19b / NFR-19b gap that left mobile users staring at empty screens or errors when the device went offline.                                                                                            
                                                                                                                                                                                                                     
  ### What it does                                                                                                                                                                                                   
                                                                                                                                                                                                                     
  - A typed read-cache (`src/cache/offlineCache.ts`) backed by `expo-file-system`'s `Paths.cache` — already a transitive dependency, no new packages added. Each entry is `{ data, cachedAt }` and reads return a    
  `fresh` flag based on a configurable TTL (default 1h).                                                                                                                                                             
  - A generic `useCachedFetch(userId, slug, fetcher)` hook (`src/hooks/useCachedFetch.ts`):                                                                                                                          
    - Reads the cache on mount and shows the cached payload immediately — no spinner if the cache is fresh.                                                                                                          
    - Fires the network fetch in parallel; success overwrites both state and disk.                                                                                                                                   
    - Errors do **not** clear cached data; the user keeps seeing the last-known-good payload.                                                                                                                        
    - Refetches on `AppState` 'active' transitions (closest signal to "online again" without taking on a NetInfo dependency).                                                                                        
    - Returns a `refresh()` callback for pull-to-refresh wiring.                                                                                                                                                     
  - A composed `useOfflineCommitments(userId)` (`src/hooks/useOfflineCommitments.ts`) that splits a single `/api/handshakes/?page_size=100` round-trip into `{ active_handshakes, joined_events }`, matching the     
  FR-19b requirement.                                                                                                                                                                                                
  - A reusable `OfflineSyncBanner` component that surfaces three states — fresh, cached, error — with a relative "last synced N minutes ago" string that auto-updates every 30s. Tap to refresh.                     
  - A new `MyCommitmentsScreen` registered under `ProfileStack` that proves the integration end-to-end: shows handshake + joined-event cards, deep-links into `ServiceDetail`, supports pull-to-refresh, and reads   
  from disk when the network is unreachable.                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                         
  Closes #322                         

  ## Files

  **Mobile**                                                                                                                                                                                                         
  - `src/cache/offlineCache.ts` — typed save/read/clear plus `formatLastSynced`
  - `src/cache/__tests__/offlineCache.test.ts` — 12 tests covering key namespacing, round-trip, TTL freshness, corrupt-entry handling, overwrite, single-key + bulk clear, and time formatting                       
  - `src/hooks/useCachedFetch.ts` — generic read-through cache hook                                                                                                                                                  
  - `src/hooks/useOfflineCommitments.ts` — composes `listHandshakes` + the cache                                                                                                                                     
  - `src/presentation/components/OfflineSyncBanner.tsx` — status banner                                                                                                                                              
  - `src/presentation/screens/MyCommitmentsScreen.tsx` — new "My commitments" screen                                                                                                                                 
  - `src/navigation/ProfileStack.tsx` — `MyCommitments` route                                                                                                                                                        
                                                                                                                                                                                                                     
  ## Test plan                                                                                                                                                                                                       
                                                                               
  - [x] `cd mobile-client && npm test -- src/cache` — 12 / 12 green                                                                                                                                                  
  - [x] `cd mobile-client && npm test` — full Jest suite green
  - [ ] Manual: log in, open Profile -> My commitments, see active handshakes and joined events with a green "synced now" banner                                                                                     
  - [ ] Manual: airplane-mode the device, kill the app, reopen, navigate to My commitments -> the cached payload renders with an amber "Cached. Last synced X mins ago" banner                                       
  - [ ] Manual: while offline, tap the banner -> a red banner shows the fetch failure but cached data stays visible                                                                                                  
  - [ ] Manual: re-enable wifi, background and foreground the app -> banner flips to green, payload refreshes via the AppState resync

### DIFF
--- a/mobile-client/package-lock.json
+++ b/mobile-client/package-lock.json
@@ -26,6 +26,7 @@
         "expo-file-system": "~19.0.21",
         "expo-image-picker": "~17.0.10",
         "expo-location": "~19.0.8",
+        "expo-network": "~8.0.8",
         "expo-notifications": "~0.32.16",
         "expo-secure-store": "~15.0.8",
         "expo-status-bar": "~3.0.9",
@@ -6747,6 +6748,16 @@
       "peerDependencies": {
         "react": "*",
         "react-native": "*"
+      }
+    },
+    "node_modules/expo-network": {
+      "version": "8.0.8",
+      "resolved": "https://registry.npmjs.org/expo-network/-/expo-network-8.0.8.tgz",
+      "integrity": "sha512-dgrL8UHAmWofqeY4UEjWskCl/RoQAM0DG6PZR8xz2WZt+6aQEboQgFRXowCfhbKZ71d16sNuKXtwBEsp2DtdNw==",
+      "license": "MIT",
+      "peerDependencies": {
+        "expo": "*",
+        "react": "*"
       }
     },
     "node_modules/expo-notifications": {
--- a/mobile-client/package.json
+++ b/mobile-client/package.json
@@ -36,6 +36,7 @@
     "expo-file-system": "~19.0.21",
     "expo-image-picker": "~17.0.10",
     "expo-location": "~19.0.8",
+    "expo-network": "~8.0.8",
     "expo-notifications": "~0.32.16",
     "expo-secure-store": "~15.0.8",
     "expo-status-bar": "~3.0.9",
@@ -71,7 +72,8 @@
       "**/__tests__/**/*.test.ts"
     ],
     "moduleNameMapper": {
-      "^expo-secure-store$": "<rootDir>/src/api/__tests__/__mocks__/expo-secure-store.ts"
+      "^expo-secure-store$": "<rootDir>/src/api/__tests__/__mocks__/expo-secure-store.ts",
+      "^expo-network$": "<rootDir>/src/api/__tests__/__mocks__/expo-network.ts"
     }
   },
   "private": true
--- a/mobile-client/src/api/__tests__/__mocks__/expo-network.ts
+++ b/mobile-client/src/api/__tests__/__mocks__/expo-network.ts
@@ -1,0 +1,45 @@
+// Jest mock for expo-network. The real module is a TurboModule so it cannot
+// be imported in Node. Tests can override the listener payload by calling
+// `__setNetworkState` to simulate online/offline transitions.
+
+let _state: { isConnected?: boolean; isInternetReachable?: boolean } = {
+  isConnected: true,
+  isInternetReachable: true,
+};
+const _listeners: Array<(s: typeof _state) => void> = [];
+
+export const NetworkStateType = {
+  NONE: "NONE",
+  UNKNOWN: "UNKNOWN",
+  CELLULAR: "CELLULAR",
+  WIFI: "WIFI",
+};
+
+export async function getNetworkStateAsync() {
+  return { ..._state, type: _state.isConnected ? NetworkStateType.WIFI : NetworkStateType.NONE };
+}
+
+export function addNetworkStateListener(cb: (s: typeof _state) => void) {
+  _listeners.push(cb);
+  return {
+    remove() {
+      const i = _listeners.indexOf(cb);
+      if (i >= 0) _listeners.splice(i, 1);
+    },
+  };
+}
+
+export function useNetworkState() {
+  return { ..._state };
+}
+
+// Test-only helpers (not part of the real expo-network surface).
+export function __setNetworkState(next: { isConnected?: boolean; isInternetReachable?: boolean }) {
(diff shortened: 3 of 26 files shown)


sgunes16:
I tested the claimed offline behavior on a physical Android device with USB ADB.

After loading the app online, I disabled both Wi-Fi and mobile data via ADB

Findings:

The app does not appear to provide the promised offline experience. Once offline, the development build immediately loses its Metro connection (`Cannot connect to Metro`, `URL: 192.168.1.13:8081`). More importantly, the actual offline cache implementation seems limited to `MyCommitmentsScreen` via `useOfflineCommitments`, `useCachedFetch`, and `offlineCache`.

That screen is registered in `ProfileStack`, but I could not find a visible navigation path to it from `ProfileScreen`, so the offline surface does not seem reachable through the normal user flow.

Other screens, such as Home, may continue showing already-mounted in-memory data after the network is disabled, but this does not look like durable offline support. Refreshing or navigating to network-backed surfaces still depends on live API/Metro connectivity.

Expected:

- Offline-supported screens should be reachable from the app UI.
- Cached data should persist after network loss and app restart.
- Offline state should be clearly communicated to the user.
- Refresh failures should preserve cached data instead of behaving like ordinary network failures.
- The feature should be tested against a real no-network state, not only an already-mounted in-memory screen.

Current conclusion: the branch contains a partial offline cache helper, but it does not yet deliver a usable offline feature through the product UI.


A broader concern is that the offline model does not appear to cache core user/session data either. In most modern mobile apps, offline support is not limited to a single feature-specific list. At minimum, the app usually keeps a local snapshot of:

- the current user profile
- auth/session metadata needed to render the signed-in shell
- recently viewed feed/list data
- recently opened detail screens
- lightweight relationship/state metadata such as active commitments or conversations

This allows the app to reopen into a useful signed-in state even when the network is unavailable. The UI can then mark the data as stale, disable actions that require the server, and retry synchronization when connectivity returns.

In this branch, the offline cache seems scoped only to active commitments, and it is not clearly integrated into the main app shell. As a result, the app behaves more like an online-only client with one partial cached helper rather than a mobile offline-capable experience.

Suggested direction:

- cache the authenticated user snapshot locally after successful login/profile refresh
- hydrate the app shell from that cached user snapshot on startup
- cache important read-only surfaces such as Home feed, Profile, Messages list, and Service detail
- show a consistent offline/stale-data banner
- disable or queue server-mutating actions while offline
- resync cached surfaces when the app returns online

For a short-term version that fits this branch, the feature could be narrowed and made explicit instead of trying to imply full offline support.

For example, when the app detects no connection, it could show a dedicated offline state:

- clear title: `You are offline`
- short explanation: `Some features need an internet connection. You can still view your last synced commitments.`
- last sync timestamp
- retry button
- below that, a `Last commitments` section rendered from the existing cached `MyCommitments` data

This would make the current implementation much more honest and usable: the user would understand that the app is offline, server actions are unavailable, but their last cached commitments are still accessible.

This would also require making the commitments cache reachable from the normal UI, for example from Profile or the offline screen itself. Without that entry point, the current cached screen is effectively hidden from users.

yusufizzetmurat:
It is now fixed, asking for review.

sgunes16:
image


Above error fixed and other things are ok it works

sgunes16 (approved):

PR #474, opened 2026-05-04
 feat(ranking,privacy,achievements): friends-of-friends helper, distance blur, in-progress achievements     

### LINKED ISSUE #167: Investigate Missing Achievement Progress in API Response
Currently, the achievements endpoint appears to return only completed achievements, while achievements that are still in progress are not included in the response.

This makes it difficult for the frontend to display achievement progress for users.

Possible causes:
- The endpoint may have been modified to filter only completed achievements.
- Progress-related fields may not be included in the serializer.
- The backend logic may not calculate or return in-progress achievements.

Tasks:
- Verify whether the achievements endpoint has changed.
- Check if the API intentionally filters only completed achievements.
- Ensure that achievements with progress are also returned.
- Confirm that the response includes progress-related fields required by the frontend.

Expected behavior:
The API should return both completed achievements and achievements with ongoing progress so the frontend can display progress indicators.

### LINKED ISSUE #318: [FR-17k] Social graph for friends-of-friends proximity boost in search
**The problem:**
The weighted search spec allocates a 0.5 weight to social proximity, services offered by people
the current user is connected to through prior transactions. There is currently no way to compute
this: the social graph is implicit in the `Handshake` table but has never been materialised.

**What is needed:**
1. `get_social_neighbours(user, depth=2)` — returns a set of user IDs reachable within 2 hops
   through completed handshakes (direct transaction partners, then their partners).
2. A boost annotation or filter step in the search pipeline that applies the 0.5 weight multiplier
   for services from users in that set.

**Design questions to resolve:**
- Maximum depth: 2 hops is friends-of-friends; 3 hops increases query cost significantly.
- Caching: social graphs change slowly — a per-user cache with a configurable TTL (e.g., 1 hour)
  is reasonable.
- Computation: at MVP scale, a 2-hop JOIN over the `Handshake` table is sufficient. At scale,
  consider a PostgreSQL recursive CTE or a pre-computed adjacency table.
- Anonymous users: boost is 0, no graph lookup needed.

**Acceptance criteria:**
- `get_social_neighbours(user, depth=2)` returns correct IDs for direct, 2-hop, and no-connection cases.
- Self-loops are guarded against (user cannot be their own neighbour).
- Result is cached per user with a configurable TTL.
- Search results include a proximity boost for authenticated users.
- The boost is 0 for anonymous requests.

### LINKED ISSUE #319: [FR-17l] Location privacy: blur distance to 500m in the feed until handshake is accepted
**The problem:**
The discovery feed currently exposes distance values computed from real coordinates. A user can
infer a provider's approximate address by issuing repeated feed queries from different reference
points (triangulation). The existing location blur in `serializers.py` applies to the service
detail view but not to distance values in the feed.

**What the design requires:**
- Distance values shown in feed responses must be rounded to the nearest 500m for services where
  no accepted handshake exists between the requester and the provider.
- PostGIS real coordinates are still used internally for proximity filtering and ranking.
- After a handshake reaches `status='accepted'`, exact distance (or address detail) is available.

**Current behaviour:**
- All non-owners always see fuzzed coordinates.
- Exact coordinates are exposed via `session_exact_location_lat/lng` fields — but these appear
  to be a separate session-scoped mechanism, not the main `location_lat/lng` fields.

**What needs to happen:**
- In the serializer's `to_representation`, check whether the requesting user has an `accepted`
  handshake with the service.
- If yes: return the real coordinates in `location_lat/lng`.
- If no: return the fuzzed coordinates as today.
- The fuzz offset should stay deterministic (current FNV-1a approach is good) so that
  repeated queries don't enable triangulation.

**Design decision required:**
- Does the reveal apply per-participant (only the matched requester sees real coords) or
  per-service-once-any-handshake-is-accepted?
- What happens on mobile — does the mobile client also receive fuzzed coordinates?

**Acceptance criteria:**
- Feed distance values are quantised to 500m multiples pre-handshake.
- Real PostGIS distance is used for the ranking/filtering computation internally.
- Post-handshake, the exact value is visible.
- Anonymous users always see the blurred value.
- A unit test for the blur function; an integration test for pre/post-handshake behaviour.

yusufizzetmurat:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
All small backend changes that share zero risk surface but are worth bundling because they're each one-file fixes.                                              
                                                                                                                                                                                                                     
  ### #318 (FR-17k) — Friends-of-friends helper                                                                                                                                                                      
                                                                                                                                                                                                                     
  Adds `get_social_neighbours(user, depth=2)` in `api/services.py` next to the existing `get_social_proximity_boosts`. Returns a flat `set[str]` of user IDs reachable in `depth` hops via the existing `UserFollow` 
  + completed `Handshake` graph. Cached per `(user_id, depth)` for an hour because the graph changes slowly.                                                                                                         
                                                                                                                                                                                                                     
  The 0.5 weight multiplier called for in the FR-17k spec is already wired in `ServiceViewSet.get_queryset` via `composite_score = hot_score + 0.5 * social_boost`, so this PR just exposes the deterministic set so 
  other consumers (search filters, tests, future CLI) can reuse the same answer without going back through the boost dict.
                                                                                                                                                                                                                     
  ### #319 (FR-17l) — Distance blurred to 500m until handshake accepted                                                                                                                                              
                                      
  `ServiceSerializer.to_representation`:                                                                                                                                                                             
  - Replaces the event-only `_is_event_participant` gate with a more general `_has_accepted_handshake` that returns true for any service the requester has an `accepted` / `completed` / `reported` / `paused`
  handshake with. Event participants (RSVP statuses) are still covered.                                                                                                                                              
  - Once `_has_accepted_handshake` is true, the existing exact-coords path applies and `session_exact_location_*` is exposed.
  - New `distance` field handling: when `LocationStrategy` annotates the queryset (lat/lng provided) the serializer now exposes the distance to the viewer. For non-handshake-partners that distance is rounded to   
  the nearest 500m (`_blur_distance_to_500m`), so repeated feed queries from different reference points cannot triangulate the provider's address. Owners and accepted-handshake partners see the precise meter      
  value.                                                                                                                                                                                                             
                                                                                                                                                                                                                     
  ### #167 — Achievement progress includes in-progress entries                                                                                                                                                       
                                      
  The endpoint already iterated all `ACHIEVEMENT_REQUIREMENTS` and returned both earned and unearned entries, but unearned entries lacked an `id` field on the value side, which is what the frontend needed to      
  render them as a list. This PR:                                              
                                                                                                                                                                                                                     
  - Echoes `id` onto each value alongside the existing dict key so consumers can flatten with `Object.values(progress)`.                                                                                             
  - Adds a convenience `in_progress` boolean (`!earned && visible && current > 0`) so the response shape can't regress to "earned only" without breaking a test.
  - Tightens the `is_visible` derivation so hidden-not-earned achievements continue to omit `current` / `threshold` (they would otherwise render as null phantoms in the UI).                                        
                                                                                                                                                                                                                     
  ### Tests                                                                                                                                                                                                          
                                                                                                                                                                                                                     
  `backend/api/tests/integration/test_ranking_privacy_achievements_api.py` — 8 tests:                                                                                                                                
  - `TestSocialNeighbours`: anonymous returns empty; 1-hop via completed handshake; 2-hop friends-of-friends inclusion + 1-hop exclusion.
  - `TestLocationBlur`: anonymous viewer's distance is rounded to 500m; handshake partner sees exact coords; non-partner sees fuzzed coords.                                                                         
  - `TestAchievementProgressEndpoint`: progress includes unearned `first-service` with `id` + `current` + `threshold`; endpoint returns the same payload with `in_progress` flag.                                    
                                                                                                                                                                                                                     
  Closes #318                                                                                                                                                                                                        
  Closes #319                                                                                                                                                                                                        
  Closes #167                                                                                                                                                                                                        
                                                                               
  ## Files                            

  - `backend/api/services.py` — `get_social_neighbours`, `SOCIAL_NEIGHBOURS_TTL`                                                                                                                                     
  - `backend/api/serializers.py` — `_has_accepted_handshake`, `_blur_distance_to_500m`, broader `show_exact` gate, distance field in `to_representation`
  - `backend/api/achievement_utils.py` — `id` + `in_progress` on every progress entry, cleaner `is_visible` derivation                                                                                               
  - `backend/api/tests/integration/test_ranking_privacy_achievements_api.py` — 8 new tests                                                                                                                           
                                                                                                                                                                                                                     
  ## Test plan                                                                                                                                                                                                       
                                                                                                                                                                                                                     
  - [ ] `cd backend && .venv/bin/python -m pytest api/tests/integration/test_ranking_privacy_achievements_api.py -v` — 8 green                                                                                       
  - [ ] `make test` — full suite green
  - [ ] Manual: log in as user A, view a service whose owner you have NO handshake with -> `location_lat` / `location_lng` are fuzzed and `distance` (when present) ends in `00`                                     
  - [ ] Manual: accept a handshake on that service as user A -> reload the service detail -> exact `location_lat` / `location_lng` plus precise `distance` value                                                     
  - [ ] Manual: `GET /api/users/{me}/badge-progress/` -> response includes `first-service` with `earned: false`, `current: 0`, `threshold: 1`, `id: "first-service"`, `in_progress: false`


mzyavuz (approved):
After the handshake, the participants informed about the location. All acceptance tests have passed. Approving it!

PR #491, opened 2026-05-05
feat(ranking): personalized For You feed with section carousel

### LINKED ISSUE #481: Personalized For You feed beyond the global hot ranking
Today every user sees the same hot list. That is fine for browse but it is not what people expect from a feed in 2026. They expect something that looks at *them*. I want to build a real For You feed that combines several viewer specific signals on top of the existing hot ranking. Overlap with their declared interests, who they follow (and who their follows follow), what services tend to appear together in completed handshakes among similar users, and a soft penalty so the feed does not keep showing them the same cards they just saw.

This is big enough that I will spec it separately before any code goes in. The follow graph piece overlaps with the FR DIS 05 social proximity work that was deferred from #458, so that gets folded in here.

Acceptance criteria:
1. A new feed surface (tab or page) is available to authenticated users, distinct from the existing hot or chronological browse.
2. The For You ranking blends viewer onboarding skills, follow graph affinity, handshake cooccurrence, and viewing recency on top of the hot ranking baseline.
3. Blend weights and the algorithm are documented in a spec reviewed before implementation.
4. Performance stays within the existing one second feed SLA.
5. A way to evaluate feed quality is in place (offline holdout, CTR proxy, or a defined metric in the spec).

yusufizzetmurat:
## Summary

A real personalized feed on top of the existing Hot ranking. New `?sort=for_you` blends four viewer-specific signals additively on `hot_score`: tag overlap with the viewer's declared skills, follow-graph affinity (reuses `get_social_proximity_boosts` and finally consumes the deferred FR-DIS-05 weight), anonymized handshake cooccurrence with a k-anonymity threshold of 3, and a recency-of-viewing decay over a 7 day window in the Django cache.

Surface is a horizontal "For you" carousel above the existing Hot feed on both the web dashboard and the mobile home screen. Anonymous and not-onboarded viewers see no carousel. Cards link with `?from=for_you` so the new `ForYouEvent` CTR proxy can attribute clicks and downstream handshakes against a Hot baseline. An admin endpoint at `/api/services/for-you-metrics/?days=N` exposes the rollup for eyeballing CTR without standing up a dashboard.

Daily management commands rebuild the cooccurrence matrix and roll up `ForYouEvent` into `ForYouDailyMetric`.

Closes #481

## Test plan

- [ ] `cd backend && .venv/bin/python -m pytest api/tests/unit/test_for_you_scorer.py api/tests/integration/test_for_you_endpoint.py` (37 new tests, plus full suite green at 1264 passed locally)
- [ ] `cd frontend && npm run lint && npx tsc --noEmit -p tsconfig.app.json && npm run test -- --run`
- [ ] `cd mobile-client && npx tsc --noEmit && npm test`
- [ ] `cd backend && .venv/bin/python manage.py build_handshake_cooccurrence` and `.venv/bin/python manage.py roll_up_for_you_metrics` both run cleanly
- [ ] `make dev`, sign in as an onboarded user with skills, confirm the For you carousel renders above the Hot feed; click a card and confirm `?from=for_you` is set in the URL; complete a handshake and check `/api/services/for-you-metrics/?days=7` as an admin shows impression, click, and handshake rows


mzyavuz (dismissed):
### Review notes

#### UX — Map collapse is too sensitive (my observation)

The 80 px scroll threshold makes the map flicker open and closed with small scroll movements, scroll down slightly to open "Try something new," then natural scroll-back collapses the map, creating a jitter loop. This is a real usability issue, not cosmetic. Two fixes should both be applied:

1. **Add hysteresis**: collapse at 80 px down, but only re-open when scroll returns to ≤ 20 px (not 80 px). This prevents the flicker band.
2. **Debounce the scroll handler**: the current handler fires on every scroll event with no debounce, which makes the toggle feel abrupt.

#### Ranking signal quality (tag match is too loose - ONLY MY OPINION/SUGGESTION)

With Math and Software skills, the first 3 results have an Education tag, the 4th has Art, and the 5th has Technology. Technology is the closest match and should rank higher. This suggests:

- The `tag_overlap()` Jaccard score is not differentiating well, likely because Education is a **parent** of both Math and Software in the Wikidata hierarchy — so those services get partial credit through `parent_qid` expansion even though they're not a direct match.
- The `TAG_WEIGHT` (0.3) may be too low relative to `FOLLOW_WEIGHT` (0.4), meaning follow-graph affinity is overriding tag relevance for new users who don't yet have many connections.

Worth flagging to the author: either increase `TAG_WEIGHT` or tighten the parent QID expansion (e.g., only expand one level up, not arbitrarily deep). A demo with a fresh onboarded user with specific skill tags would surface this clearly in the test plan.

#### Metrics endpoint (`count: 280` is easy to misread)

The admin endpoint returns a raw impression count (one row per card per viewer per load), not a unique-viewer count or page-view count. Without documentation, this number looks inflated and will confuse anyone reading the admin panel. The endpoint response should include a `note` field or the API docs should clarify: `"count": 280` means 280 card impressions, not 280 users or 280 page views. Adding `unique_viewers` (a `COUNT(DISTINCT viewer_id)`) as a second field would make the data immediately more useful and prevent misreads.



### Summary of Code Review Issues 

#### Backend — Blocking Before Merge

**1. Cooccurrence matrix rebuild is not atomic**
If the process crashes during `rebuild_cooccurrence_matrix()`, the table is left empty (since it's deleted before bulk_create completes). This silently breaks the cooccurrence signal for all users until the next nightly run. Wrap the delete + insert in a transaction:
```python
with transaction.atomic():
    HandshakeCooccurrence.objects.all().delete()
    HandshakeCooccurrence.objects.bulk_create(...)
```

---

#### Backend — Should Fix Soon

**2. Handshake attribution fires on every Exchange update, not just completion**
The `post_save` signal runs on PENDING → ACCEPTED → COMPLETED transitions, but only the COMPLETED transition is meaningful for CTR. Add an early exit:
```python
if not created and instance.state != Exchange.COMPLETED:
    return
```

**3. Impression cache must be Redis-backed for multi-worker deployments**
The recency penalty uses Django's cache to track per-viewer impression history. In development (single Daphne worker) this works fine, but at scale with multiple workers, each worker has its own cache and the recency signal diverges. Verify the implementation uses `django.core.cache.cache` (Redis) and not a local dict. Add a comment confirming this assumption.

**4. `ForYouEvent` table has no cleanup strategy**
The append-only event log grows unbounded. After rolling up to `ForYouDailyMetric`, the raw `ForYouEvent` rows should be pruned. Add a `--prune-days=N` flag to `roll_up_for_you_metrics` to delete rows older than N days.

**5. Admin metrics endpoint can double-count if rollup runs intraday**
If `roll_up_for_you_metrics` is run twice in a day, "today's" rolled metric might include data already in the live query. The unique constraint prevents duplicates in the rolled table, but this edge case should be documented in the endpoint docstring.

**6. Serializer fields are transient (set by view, not model)**
The new `source`, `for_you_signals`, `explore_pool` fields are populated dynamically by `_list_for_you()` before serialization. This works but is non-standard. Add inline comments to each field in the serializer so future readers understand they're not stored on the model.

---

#### Frontend — Nice-to-Have

**7. `ForYouCarousel` and `ExploreCarousel` swallow errors silently**
Network failures or API 500s set `items = []` with no logging. Add a console.error to aid debugging.

**8. Carousel diversification has a hardcoded lookahead**
`diversifyByChip()` uses `LOOKAHEAD = 3` which works fine for a 10-card carousel but could produce different visual results on mobile (5 cards). Add a comment noting the assumption.

**9. Map collapse threshold lacks test coverage**
The 80 px scroll threshold in `DashboardPage` is not tested. Good to document in the component so it's not accidentally removed.

**10. Mobile has no rendering tests**
`ForYouSection` TypeScript types are checked, but no snapshot or rendering tests exist. Acceptable for MVP but worth flagging.

---

#### Tests

**Missing:** No test for cooccurrence rebuild atomicity (what happens if the process crashes mid-rebuild). Given the delete-then-insert pattern, a test confirming the old data survives a thrown exception would catch regressions.

yusufizzetmurat:
Blocking items are fixed and added comments on the code and test.

sgunes16:
approved but ui fix needed

sgunes16 (approved):

sgunes16 (approved):

Issue #483, opened 2026-05-05
Reshare, endorse, save mechanics for services and providers


yusufizzetmurat:
Once the activity feed exists, we want users to push signal into it. Endorse a provider publicly, save services to private lists.

Holding off on speccing the details until E1 through E3 are in flight, since the surface area depends on what shape the activity feed takes.

Acceptance criteria:
1. Users can publicly endorse a service provider, with endorsements visible on the provider profile and counted as a signal.
2. Users can save services to private personal lists, viewable only by themselves.

PR #493, opened 2026-05-05
feat(activity): chronological activity feed of follows and nearby people

### LINKED ISSUE #482: Activity feed of follows and people nearby
The platform does not have any social timeline today. There is no surface that says "someone you follow posted a new service", "an event in your area is filling up", "your friend joined this group offer". That is the social spine, and without it the platform feels like a directory rather than a network.

I want a basic activity feed. A chronological timeline of events from people the user follows plus events from their broader area. Cards are mostly references back into the existing service, handshake, and event surfaces. The feed itself is just the index. Will need a small data model for activity events and producers wired into the existing signals.

Like For You, this gets its own spec before code.

Acceptance criteria:
1. A new endpoint returns a chronological list of activity events from followed users and from users within the configured proximity radius of the viewer.
2. Activity events are produced via existing signal hooks when services are created, handshakes accepted, events RSVPd, and follows created.
3. The frontend renders the feed as a card list, with each card linking back into the relevant service, handshake, or event surface.
4. Privacy controls are in place for event participation visibility (defaults defined in the spec).
5. A spec document is written and reviewed before implementation begins.

yusufizzetmurat:
## Summary

A chronological activity feed so the platform feels less like a directory and more like a network. New `ActivityEvent` model captures three verbs (`service_created`, `handshake_accepted`, `user_followed`) emitted by signal handlers when those things happen. `GET /api/activity/feed/?days=N&lat=...&lng=...` returns events from people the viewer follows plus actors within `RANKING_PROXIMITY_HALF_LIFE_KM*3` of the viewer's location. Viewers without a known location only see events from people they follow; the viewer's own events are excluded.

Web mounts `/activity` as a route with a top nav entry. Mobile adds an `ActivityScreen` in the HomeStack reachable from a pulse icon on the home header. Both render the same chronological card list with cards linking back into the relevant service.

Hidden services produce no events. The verb taxonomy stays small for v1; reshare and endorse arrive in #483.

Closes #482

## Test plan

- [x] `cd backend && .venv/bin/python -m pytest api/tests/integration/test_activity_feed.py api/tests/unit/test_signals.py api/tests/integration/test_handshake_api.py api/tests/integration/test_user_follow_views.py` (1236 backend tests pass locally with no regressions)
- [x] `cd frontend && npm run lint && npx tsc --noEmit -p tsconfig.app.json && npm run test -- --run`
- [ ] `cd mobile-client && npx tsc --noEmit && npm test`
- [x] `make dev`, follow another user, post a service from that other account, confirm `/activity` shows the event with a working link to the service
- [ ] Pass `lat=...&lng=...` and confirm events from nearby actors (services with location) also appear without needing a follow


mzyavuz (changes requested):
The activity feed idea is great, I love it, it shows when someone you follow posts a service or when you get a new follower, which makes the platform feel much more alive. The backend structure looks solid overall too.

### Things I noticed while testing

**Location prompt showing even when location is already enabled**

When I opened the activity page, the right rail was showing the "Enable your location" prompt even though my location was already enabled. So something isn't reading the location state correctly. Also because of this, the right side of the page is basically just that one prompt with a ton of empty space below it, looks a bit unfinished.

**Following and Nearby filter counts both show 0**

The filter chips show "Following 0" and "Nearby 0" even though there are clearly services from people I follow in the feed. I think this is related to the location issue above, if the location isn't being passed to the API properly, distance_km would always be null and the nearby filter would always be empty. But the following count being 0 is also strange on its own.

**UI needs some work & tagging @citizenduck (Selman Güneş)**

The page works but the design feels a bit off. The hero card images are long and narrow which looks awkward, and the "Open" button being fully black is a bit heavy. Selman, can you take a look?

**Follow activity not reflected on the followed user's side**

When a new verified user follows another verified user, the old user doesn't see that follow in their activity feed. So the user_followed event either isn't being emitted correctly or isn't showing up for the target user, tested and confirmed from the verified user's account.

**No notifications for new follows or activity events**

There are no notifications showing up in the notifications page for new follows or activity events. So even if the feed was working, users wouldn't get notified when someone follows them or when activity happens from people they follow.

**Find people to follow" page is missing**
                                                                                         
When the feed is empty or when I tried to find people to follow to actually populate the feed, there was no "Find people to follow" page or discovery flow. This makes it hard to even test the feed properly, if there's no way to find and follow people, the feed stays empty for most users.  

### Code things

**`_status_changed` flag might never be set (important)**

The handshake signals check `getattr(instance, '_status_changed', False)` before emitting events, but I don't see anywhere in this diff that actually sets that flag to `True`. The tests pass because they call `.save()` directly and probably hit a different path. If this flag is never set in the actual views/services flow, handshake_accepted and handshake_completed events would silently never fire. Worth double checking.

**N+1 queries in the serializer**

`get_event_capacity_pct` and `get_handshake_duration_hours` both hit the DB once per card in the list. Same pattern as `_distance_km` is already doing — would be cleaner to annotate or precompute these in `get_queryset()` or `list()` instead.

**"Follow" button in the welcome card doesn't actually follow anyone**

Both "Say hi" and "Follow" in `ActivityWelcomeCard.tsx` go to the same profile route. The Follow button should call the follow API or just be removed for now.

**Missing newlines at end of file**

Two places in `views.py` are missing a trailing newline.

**Inline `<style>` in `ActivityHeroCard`**

The `@keyframes activityPulse` animation is injected as an inline `<style>` tag inside the component, which means it gets duplicated for every hero card on the page. Should be moved to a global stylesheet.

Overall the backend is well done and the concept is great. Main things to fix before merge: the location detection bug, the `_status_changed` flag, and the UI feedback from Selman.

yusufizzetmurat:
Thanks for the review,  the changes are made, for ,e _status_changed — it's a @property on the Handshake model that compares __original_status (captured in __init__) with the current status at  save time, not a flag that needs to be set anywhere. The accept/deny/complete views all do instance.status = 'X'; instance.save(), which is the  path the test exercises. Verified the test passes against the actual view path (Handshake.objects.get(...) → status = ... → save() is the same    shape views.py:3173-3174 uses). No fix needed there.

mzyavuz (dismissed):
This is a meaningful improvement over the first version. The follow notification fix and the suggested users page were the two biggest gaps and both are addressed.

After resolving the conflicts it can be merged, I am approving it!

mzyavuz (approved):
Conflicts have been resolved. It can be merged now. Approving it!

PR #494, opened 2026-05-05
feat(social): save and endorse mechanics for services

### LINKED ISSUE #483: Reshare, endorse, save mechanics for services and providers

yusufizzetmurat:
## Summary

**Save** is a private bookmark. `SavedService` model with unique `(user, service)`. `POST /api/services/<id>/save/` toggles it on, `DELETE` removes it, `GET /api/services/saved/` lists the viewer's saves. Service serializer carries `is_saved` per viewer. Web adds a Saved page (linked from the user dropdown). Mobile and web both get a Save button on the service detail surface.

**Endorse** is a public vouch for the service's provider. `Endorsement` model with unique `(endorser, service)`. Same toggle pattern: `POST/DELETE /api/services/<id>/endorse/`. Self-endorsement is rejected with 400. Service serializer carries `is_endorsed` (per viewer) and `endorsement_count` (public). Endorsements are intentionally not yet wired into Wilson quality so this PR keeps the ranking math untouched; the ranking integration is a planned follow-up.

Closes #483

## Test plan

- [x] `cd backend && .venv/bin/python -m pytest api/tests/integration/test_save_endorse.py api/tests/integration/test_service_api.py api/tests/unit/test_signals.py` (62 pass locally, no regressions)
- [x] `cd frontend && npm run lint && npx tsc --noEmit -p tsconfig.app.json && npm run test -- --run`
- [x] `cd mobile-client && npx tsc --noEmit && npm test`
- [x] `make dev`, on a service detail page tap Save, refresh, confirm the button shows "Saved" and the service appears under `/saved`
- [x] Tap Endorse on a service you don't own; confirm the count increments and the button toggles. Try to endorse your own service and confirm the API responds 400


mzyavuz (dismissed):
Reviewed the PR and ran all the acceptance tests manually (save, unsave, endorse, unendorse, self-endorse block (got the 400 as expected)), and the saved list showing only my own bookmarks. All passed. 

One thing to fix before merge: in the `saved` action in `views.py`, you're setting `svc.is_saved = True` directly on the model instance, but the serializer reads from `is_saved_anno` (not `is_saved`). So the annotation fast path is never hit and it falls back to an extra `.exists()` query per service in the saved list. Should be `svc.is_saved_anno = True` instead.

Good to go after the conflict resolution.

yaseminsirin (approved):
Tested the main save/endorse flows on web and reviewed the API/UI behavior — everything looks good from my side. The separation between private saves and public endorsements is clean, the toggle behavior feels consistent, and the self-endorsement restriction works as expected.

Nice addition overall, approved 👏

PR #500, opened 2026-05-06
fix(notifications): fix notification routing

### LINKED ISSUE #496: Event notifications do not redirect to the events page
## Problem
Event notifications are currently directing users to the chat page instead of the events page. This is confusing since the event-related chats are actually listed under the events page.

## Expected Behavior
When an event notification is clicked, the user should be redirected to the corresponding event on the events page.

## Actual Behavior
Clicking an event notification takes the user to the chat page. Navigating to the events page is then required to find event chats.

## Steps to Reproduce
1. Trigger an event notification.
2. Click on the notification.
3. Observe that you are directed to the chat page instead of the events page.

## Suggested Solution
Update the notification redirection logic so that event notifications reliably direct users to the appropriate event entry within the events page, ensuring event chats remain accessible as intended.

### LINKED ISSUE #498: Admin Notifications are Not Linked to Detail Page
**Where:**
Admin notification list (`/NotificationsPage.tsx`) or admin panel

**Problem:**
Admin notifications do not navigate to the relevant detail page when clicked. For example, a report notification should open the corresponding report detail page, but currently it does not.

**Expected:**
Each notification should deep-link to the correct entity detail page. When an admin clicks on a notification, it should navigate to the associated item's detail page such as a report, user profile, submission, etc.

### LINKED ISSUE #499: Missing Notifications Context/Detail
**Problem:**
Notifications display only a generic message without enough detail to understand what triggered them. Additionally, it should be verified that each notification type is wired to the correct target URL/screen.

**Expected:**
Each notification should show relevant context (e.g., service title, user name) and route to the correct detail view.

### LINKED ISSUE #517: Bug: Users receive notifications for their own chat messages
Users currently receive notifications for chat messages that they sent themselves. This creates noisy and confusing notification behavior, especially in active one-to-one, group, or event chat threads.

## Current Behavior

- A user sends a message in a chat.
- The notification system creates or delivers a notification back to the same sender.
- The sender may see an unread notification or push/in-app notification for their own message.

## Expected Behavior

- Message notifications should be delivered only to other participants in the conversation.
- The message sender should never receive a notification for their own message.
- Existing notification behavior for recipients should remain unchanged.

## Scope

- Review notification creation and delivery logic for:
  - one-to-one chat messages
  - group chat messages
  - event/public chat messages, if applicable
- Ensure sender exclusion is enforced on the backend, not only hidden in the frontend.
- Keep notification read/unread behavior unchanged for valid recipients.

## Tasks

- Identify where chat-message notifications are created and broadcast.
- Exclude the message sender from notification recipient lists.
- Verify WebSocket/push notification payloads are not sent back to the sender.
- Add backend regression tests for sender-exclusion behavior.
- Add or update mobile/web tests if notification list behavior is covered there.

## Acceptance Criteria

- When User A sends a direct chat message to User B, only User B receives a notification.
- When a user sends a group chat message, all other group participants receive a notification, but the sender does not.
- When a user sends an event/public chat message, the sender is excluded from any generated message notification.
- No unread notification is created for the sender's own message.
- Existing recipient notifications still include the correct chat/message metadata for navigation.

## Test Plan

- Backend: add integration coverage for direct, group, and event/public chat notification creation.
- Backend: assert no `Notification` row is created where `recipient == message.sender`.
- Mobile/web: manually verify that sending a message does not increment the sender's notification badge.
- Mobile/web: verify another participant still receives the notification and can open the correct chat.

Image

mzyavuz:
## Summary                                                
                                                                                                                                    
  Closes #496,
  Closes #498,
  Closes #499,
  Closes #517,                                                                     
                                                                                                                                    
  Fixes a group of related notification bugs where event notifications (RSVP, check-in, attendance confirmed, leave event) were routing users to the messages/chat screen instead of the event's service-detail page. The root cause was that `NotificationSerializer` did not expose `related_service_type`, making it impossible for clients to distinguish event notifications from regular service notifications.                                                                               

  Several additional notification bugs discovered and fixed in the same pass.                                                       
   
  ## Changes                                                                                                                        
                                                            
  - **Backend:** Added `related_service_type` as a `SerializerMethodField` to `NotificationSerializer` and added  `select_related('related_service', 'related_report')` to `NotificationViewSet.get_queryset()` to avoid N+1 queries              
  - **Backend:** Added `related_report` FK to `Notification` model (with migration) so report notifications can deep-link directly to the report in the admin panel                                                                                                  
  - **Backend:** Added `related_service_type` and `related_report` to Expo push notification payload so cold-start navigation works correctly on mobile                                                                                                               
  - **Backend:** Fixed all 5 admin notification filter queries from `role='admin'` to `role__in=['admin', 'super_admin']` so super admins receive report-related notifications                                                                                       
  - **Backend:** Added `service=handshake.service` to `positive_rep` "Leave Feedback" notifications in `confirm_completion` so they route to service detail, not chat                                                                                                 
  - **Frontend:** Added `related_service_type` and `related_report` to the `Notification` TypeScript interface                    
  - **Frontend:** Updated `handleClick` in `NotificationDropdown` and `NotificationsPage` with correct routing priority: report → event service-detail → positive_rep service-detail → handshake chat → service-detail → /notifications                             
  - **Frontend:** Fixed chat message notification URL from `?handshake=X` (query param) to `/messages/:handshakeId` (path param) to match the router                                                                                                                  
  - **Frontend:** Added ID-based deduplication to `addNotification` in the Zustand store and synced unread count via `fetchUnreadCount()` after each WebSocket message to prevent duplicate notifications and badge count drift                        
  - **Frontend:** Fixed `NotificationsPage` not scrolling (missing internal scroll container for full-screen prefix pages)        
  - **Frontend:** Removed `!compact` guard from `NotificationItem` so message body is always shown in the dropdown                  
  - **Mobile:** Added `related_service_type` and `related_report` to the `Notification` interface                                   
  - **Mobile:** Updated `navigateToNotificationTarget` to check `related_service_type === 'Event'` before the handshake check, routing event notifications to `ServiceDetail` instead of `Chat`                                                                  
                                                                                                                                    
## How to Test                                                                                                                    
                                                                                                                                  
  1. Create an event and have another user RSVP → organizer receives notification → tap it → should land on the event's service-detail page, not messages
  2. As organizer, accept the RSVP (check-in) → participant receives notification → tap it → should land on event service-detail    
  3. Have organizer cancel the RSVP → participant receives notification → tap it → should land on event service-detail              
  4. Complete a regular offer exchange → both parties receive "Leave Feedback" notification → tap it → should land on service-detail, not chat                                                                                                          
  5. Submit a report → admin and **super admin** both receive a notification → tap it → should open                                 `/admin?tab=reports&reportId=<uuid>`                                                                                              
  6. Trigger a chat message notification → tap it → should open the specific conversation, not the general messages list 
  7. Verify no duplicate notifications appear in the dropdown on first load                                                         
  8. Verify the Notifications page scrolls when the list is long


yaseminsirin (dismissed):
Image

Thanks for the work here — overall the direction looks good, but I noticed a couple of notification-flow issues during testing that should probably be addressed before merge.
Clicking New Event RSVP currently redirects to the Messages page, but no conversation is selected/opened, so the navigation feels broken/confusing from a user perspective.
New follower notifications are not clickable at the moment.

yusufizzetmurat (changes requested):
Nice fix overall, the event routing flow looks much cleaner with `related_service_type` doing the disambiguation. Two things worth addressing before merge:
                                                                                                                                                              
  ### 1. `user_followed` push notifications dead-tap on mobile                                                                                                                                                       
                                                              
  In `backend/api/utils.py`, `_send_push_notification` builds the Expo `push_data` payload with `related_service_type` and `related_report` but doesn't include `related_user`:                                      
                                                                                                                                                                               
  ```python                                                                                                                                                                                                          
  push_data = {
      'type': notification.type,                                                                                                                                                                                     
      'notification_id': str(notification.id),                            
      'related_handshake': ...,               
      'related_service': ...,  
      'related_service_type': ...,
      'related_report': ...,      
  }                 
```                                                                                                                                                                                           
   
  On the mobile side, navigateToNotificationTarget requires related_user to be truthy for the user_followed branch — so when someone taps a follow notification from the system tray, nothing happens. WebSocket     
  delivery is fine because _broadcast_notification uses the full NotificationSerializer, which is probably why this didn't show up in dev.                                                                           
         
  Could you add related_user to push_data and pull it through in usePushNotifications alongside the other new fields?                                                                                                
                                                                          
  2. NotificationsPage.handleClick lost its else { navigate('/notifications') } fallback                                                                                                                             
  
NotificationDropdown.tsx still has the final else branch but NotificationsPage.tsx no longer does. Looks accidental since the two handleClick bodies are otherwise identical. The practical impact: any notification that doesn't match a branch (old user_followed rows with related_user = null, or any future type added without updating the chain) will mark-as-read on the page but not navigate, while the dropdown still routes correctly.                                                                                                                                                                                            
                                                                          
  Easiest fix is just to re-add the else { navigate('/notifications') } to NotificationsPage to keep them in sync.

yusufizzetmurat (approved):

PR #515, opened 2026-05-07
chore(mobile): drop unused Firebase messaging and clean up dev/env setup

### LINKED ISSUE #514: chore(mobile): drop unused Firebase messaging and clean up dev/env setup
Mobile push notifications go through Expo end-to-end — the client uses expo-notifications to get an ExponentPushToken and the backend sends through exponent-server-sdk. But we also have @react-native-firebase/messaging installed and a setBackgroundMessageHandler in index.ts that never actually fires, because the payload arrives via Expo's transport, not directly through FCM. So it's dead code that bloats the build and makes the setup look like it does more than it does.

While I'm in there, a few other things make mobile painful for anyone new:

- A fresh clone with no .env silently falls back to the production API URL in app.config.ts, so the app "works" but is hitting prod.
- start:local:ios hardcodes http://localhost/api, which assumes the nginx Docker stack is running on port 80. Nothing tells you that.
- google-services.json / GoogleService-Info.plist are gitignored (correct), but there's no documented path for a new contributor to get them.
- Some screens have decorative comment separators that don't add anything.

What I want to do in one PR:

- Remove @react-native-firebase/messaging and @react-native-firebase/app from package.json and app.json plugins
- Delete the dead setBackgroundMessageHandler block in index.ts
- Remove the silent prod-URL fallback in app.config.ts; throw a clear error if EXPO_PUBLIC_API_URL is missing for non-prod profiles
- Point start:local:ios at the same backend the rest of the stack uses (localhost:8000 direct, no nginx) and document the choice
- Drop the VITE_MAPBOX_TOKEN cross-project fallback
- Add a short "running mobile locally" section to mobile-client/README.md covering env, Firebase credential request flow, and the two start scripts
- Tidy decorative comment separators in ServiceDetailScreen and MapScreen

No API or behaviour changes for end users.

yusufizzetmurat:
Closes #514.

Push notifications already go through Expo end-to-end, so `@react-native-firebase/messaging` and the `setBackgroundMessageHandler` in `index.ts` were dead weight — Expo's transport never reaches that handler. Removed both packages and the handler.

While in there, cleaned up a few things that make mobile painful for new contributors:

- `app.config.ts` no longer falls back to the prod API URL when nothing is set. It throws with a hint instead, so a fresh clone can't silently hit production.
- `start:local:ios` and `build:ios` now point at `localhost:8000` directly. The old `localhost/api` assumed nginx on port 80, which wasn't documented anywhere.
- Dropped the `VITE_MAPBOX_TOKEN` cross-project fallback in `app.config.ts`.
- Added a `mobile-client/README.md` covering the local-dev quickstart, Firebase credential flow, and how to run the tests.
- Stripped the decorative `──` comment separators from `mobile-client/src` (second commit) so future diffs stay readable.

No runtime/API changes. Push delivery is unaffected — Expo's push service still bridges to FCM/APNs under the hood.

## Test plan

- [x] `npm test` in `mobile-client/` (268 passed)
- [ ] `npm run start:local:ios` against `make dev` — token registers, foreground notification arrives
- [ ] `npm run start:local:android` on emulator — same check
- [ ] Background notification with app killed — OS banner appears, tap routes correctly
- [ ] Move `.env` aside and run `npm run start` — confirm it errors cleanly instead of pointing at prod


sgunes16:
image

image

Could not build 

encountering issue with npm run start:local:android also

yusufizzetmurat:
If you wait it should open up regardless more robust fixes are coming up

mobile-client/.gitignore
sgunes16, line 1: we can hold it anyways. Because already we have this files in our local and we may acciedentally push them to branches

sgunes16:
The build is okay right now. but we have a little issue 

Also, when opening a chat from a notification, the top bar currently shows the other user's name instead of the service/chat context, and there is no back button (some times visible sometimes not). The notification deep-link flow should open the chat inside a proper stack context(correct service) so users can navigate back, and the header should match the normal chat screen behavior.

@mzyavuz, I think you also fix this issue right?


image
image

mzyavuz:
> The build is okay right now. but we have a little issue
> 
> Also, when opening a chat from a notification, the top bar currently shows the other user's name instead of the service/chat context, and there is no back button (some times visible sometimes not). The notification deep-link flow should open the chat inside a proper stack context(correct service) so users can navigate back, and the header should match the normal chat screen behavior.
> 
> @mzyavuz, I think you also fix this issue right?
> 
> image image

I open a new issue for it, if it is the only problem than I think this PR can be approved.

sgunes16 (approved):

PR #534, opened 2026-05-07
fix(migrations): update CheckConstraint to use 'condition' instead of…

### LINKED ISSUE #533: setup-demo script fix
One of the scripts we are using to setup the dev  environment called with "make setup-demo" gives error when running migrations because of an old keyword.
Django 5.2 dropped the old CheckConstraint(..., check=...) argument. Constraints must use condition= (as in your models.py and newer migrations). Two older migrations still used check=, so loading them raised TypeError and make setup retried as if the DB were down.
This should be fixed

diclenaz7:
… 'check' for timebank balance and service constraints


yusufizzetmurat (approved):
Approved.

PR #519, opened 2026-05-07
test: testing overhaul (foundations)

yusufizzetmurat:
Testing overhaul: assertion helpers, mutation gates, perf gates, a11y baseline, requirement-mapped feature-11 E2E suite, CI workflow split, and mobile component-test scaffolding.

## What's in

**Backend**
- Body-shape assertion helpers (`assert_api_response`, `assert_problem_detail`) and a sweep of 82 status-only assertions in `test_chat_api.py` + `test_reputation_api.py`.
- WebSocket test harness with cookie + `?token=` parity (`api/tests/helpers/ws.py`).
- Freezegun time helper (`api/tests/helpers/time.py`).
- Property tests: ranking score monotonicity, time-bank net-flow, handshake state machine.
- New JWT lifecycle integration spec (login cookies, refresh rotation, expired-token rejection, logout clearing).
- Branch coverage on, pytest-socket loopback policy, in-memory channel layer enforcement for unit tests.
- Migrated 11 unittest TestCase classes (in `test_capacity_and_agreed.py` and `test_search_filters.py`) to pytest.

**Frontend**
- Vitest coverage thresholds (set just above the current baseline; ratchet up as new specs land).
- Playwright config: `fullyParallel: true`, 4 workers, 10s `expect.timeout`, dedicated `chromium-mobile` project for a11y / responsive specs.
- Stryker mutation testing on `services/api.ts`, `services/conversationAPI.ts`, `utils/dateTime.ts`, `utils/eventUtils.ts`, `utils/cookies.ts`. Targeted unit specs landed for each — local mutation score 60.2% across the set, dateTime.ts 60.5%, utils/ 82.9%.
- axe baseline on 9 high-traffic pages (`tests/e2e/a11y/baseline.spec.ts`).
- k6 perf gates: feed 1s (FR-17a), feed-auth 2s (NFR-19a), event create 1.5s (NFR-11a).
- Feature-11 requirement-mapped E2E suite (FR-11a..d) with helpers/feature11.ts, plus the `coverage-matrix.md` mapping every spec to its FR/NFR id.
- `waitForApi` / `waitForUI` / `waitForIdle` helpers replace 13 `page.waitForTimeout` calls; rule documented in `TEST_GUIDE.md`.
- Empty/loading/error state coverage on `MultiUseDetailsModal`, `AdminActivityFeed`, `NotificationsPage`, `FollowListModal`, `ActivityFilterChips`.
[...]

PR #536, opened 2026-05-08
test: finish #461 — sweep, migrate, fix flakes, mutmut bootstrap

### LINKED ISSUE #461: E2E Test Coverage Control
The project has a growing Playwright E2E suite across multiple SRS features. As the suite expands, it is difficult to quickly verify which requirements are covered, which tests are failing because of product gaps, and which areas still need new or stabilized E2E coverage.

This issue aims to introduce a clear E2E coverage control pass so the team can track requirement-level coverage, identify missing scenarios, and keep the E2E suite useful for milestone validation.

## Scope

- Review the current frontend E2E test structure under `frontend/tests/e2e/`.
- Map E2E specs to the relevant SRS features and FR/NFR identifiers.
- Identify missing, skipped, flaky, or outdated E2E tests.
- Separate product/runtime gaps from test-side selector, setup, or data issues.
- Produce a concise coverage status summary that can be used in milestone reporting.

## Tasks

- [ ] List all existing E2E spec files and the requirement IDs they target.
- [ ] Check whether each target requirement has at least one meaningful E2E scenario.
- [ ] Record skipped tests and explain whether they are blocked by missing product UI, backend behavior, or test setup.
- [ ] Re-run the E2E suite with retries disabled and capture pass/fail/skip counts.
- [ ] Triage failing tests into:
  - product/runtime gaps,
  - stale test assumptions,
  - flaky infrastructure or seeded-data issues.
- [ ] Update or create missing E2E tests for high-priority uncovered requirements.
- [ ] Add a short coverage-control summary to the relevant report or test documentation.

## Acceptance Criteria

- Each E2E-covered feature has a visible mapping between spec files and SRS requirement IDs.
- Missing E2E coverage is explicitly listed with priority and reason.
- Skipped tests include a short explanation and a clear unblock condition.
- Failing tests are categorized so the team can distinguish implementation defects from test maintenance work.
- The final summary includes total E2E pass/fail/skip counts from a no-retry run.
- The coverage-control output is suitable for milestone review or QA handoff.

## Tests

- Run the full E2E suite without retries:

```bash
cd frontend
PLAYWRIGHT_BASE_URL=http://localhost:5173 npx playwright test --retries=0 tests/e2e
```

- Run targeted suites after any stabilization work:

```bash
cd frontend
PLAYWRIGHT_BASE_URL=http://localhost:5173 npx playwright test --retries=0 tests/e2e/feature-<number>
```

- If report artifacts are needed, generate and attach Playwright HTML/JUnit outputs to the project test report directory.

yusufizzetmurat:
Follow-up to #519 — closes out the rest of the testing-overhaul work tracked in #461.

The bulk of it is mechanical: same regex/AST sweep as before, this time across the whole integration suite, plus the unittest→pytest migration for the unit files that #519 hadn't gotten to. Two CI flakes get root-causes (one was a real bug in `register_calendar_cache_key`'s retry loop, the other was ServiceFactory's iterator alternating schedule_type between consecutive test runs). Frontend gets specs for the 16 zero-coverage modules, Stryker's perTest masking is fixed, and the `assert-sweep` guardrail goes in so this debt can't quietly grow back.

What's in here:

- **Codemods** under `backend/scripts/codemods/` (libcst, with their own unit tests). One sweeps status-only `assert response.status_code == X` lines into `assert_api_response` / `assert_problem_detail`, folding trivial body-shape follow-ups into `contains=` / `schema=` kwargs (cap N=3, no fold for 4xx since `assert_problem_detail` doesn't take those). The other drops `class Foo(TestCase)` wrappers, translates `self.assertX` → `assert`, rewrites `with self.assertRaises` → `with pytest.raises`, and adds `@pytest.mark.django_db` where needed.
- **Backend assertion sweep** across the integration suite. 35 files, ~414 sites.
- **Backend unittest→pytest migration** for 13 unit test files. Per-file `env` SimpleNamespace fixtures (or split per-entity fixtures where it reads cleaner). Zero TestCase subclasses survive in `api/tests/unit/`.
- **Calendar cache flake** (`register_calendar_cache_key`) — the retry loop had a dead `return`, fixed with verify-then-exit. Three new unit tests including a monkeypatched racing-clobber scenario.
- **Group-chat WS flake** — two complementary fixes: widen the in-memory channel layer fixture to all `websocket`-marked tests, and pin `schedule_type='One-Time'` on the two TestGroupChatConsumer fixtures (the actual root cause was the factory iterator alternating, not Redis).
- **Mutmut Django bootstrap** — adds `api` to `also_copy` so the package is mirrored intact under `mutants/`, plus a top-level `backend/conftest.py` that loads .env and falls back to local-dev DB defaults. The original `AUTH_USER_MODEL` blocker is gone; the next-stage `RuntimeError: context has already been set` is documented inline in `pyproject.toml` for follow-up.
- **Frontend**: 12 new service specs + 4 new store specs, 239 cases total. `useTourStore` needs a Storage shim because jsdom 28 here ships a plain-Object localStorage without the methods. Coverage thresholds raised 11/8/10/9 → 15/14/14/11.
- **Stryker** — `coverageAnalysis: 'perTest' → 'all'` (perTest was masking survivors when tests dynamically import in beforeEach), json reporter wired for survivor extraction. Local mutation score 60.12%, passes the 60% break threshold.
- **Guardrail** — `make test-assert-sweep` and a 4-min `assert-sweep` job in `ci-backend.yml` that runs the codemod in `--check` mode and forbids new TestCase subclasses.

## Test plan

- [ ] `make test` green (756 unit + 742 integration locally)
- [ ] `make test-assert-sweep` clean
- [ ] `cd frontend && npm run test:coverage` passes the new 15/14/14/11 gate
- [ ] `cd frontend && npx stryker run` ≥ 60% mutation score
- [ ] `make test-mutation` — backend mutmut runs through stats collection (the `context has already been set` follow-up is acknowledged separately)
- [ ] CI `assert-sweep` job green on this PR

Closes #461

---

## Round 2 — hook-test precedent + mutation-deepening (commits 86dc8e8, 870085f)

Two more things landed on this branch after the round-1 review:

1. **First `renderHook` tests in the repo.** All 5 hooks under `src/hooks/` had zero coverage and there was no `renderHook` precedent anywhere — any future hook extraction had to invent the pattern from scratch. New: a tiny `mockWebSocket` helper, a short `helpers/README.md` documenting the pattern, and 49 tests across the 5 hooks.
2. **Mutation-deepening on the 5 already-mutated files.** Same files Stryker was already pointing at; the goal was to make the existing tests actually catch bugs instead of just hitting lines. Net: total mutation **60.12 → 71.38** (covered **77.43 → 89.58**), and 4 of the 5 files are now at 100% on covered code.

No production code changes outside two `// Stryker disable next-line` comments where the mutant was genuinely equivalent (with the reason inline). No threshold ratchets, no Stryker scope expansion.

### Coverage (round 2 deltas)

| Area | Before | After |
|---|---|---|
| `src/hooks/` (lines / fns) | 0 / 0 | 96.84 / 95.12 |
| Overall lines | 16.67 | 18.76 |
| Overall functions | 14.96 | 16.29 |
| Overall branches | 12.15 | 13.20 |

All four `vite.config.ts` thresholds (15 / 14 / 14 / 11) still pass. Bumping them to the new floor is a follow-up.

### Mutation (round 2 deltas, `break: 60`, `high: 80`)

| File | Before (total / covered) | After (total / covered) | Survivors |
|---|---|---|---|
| `services/api.ts` | 24.62 / 50.00 | 30.77 / 59.70 | 27 |
| `services/conversationAPI.ts` | 86.67 / 92.86 | **97.78 / 100** | 0 |
| `utils/cookies.ts` | 93.33 / 93.33 | **100 / 100** | 0 |
| `utils/dateTime.ts` | 60.53 / 67.65 | **94.44 / 100** | 0 |
| `utils/eventUtils.ts` | 88.35 / 89.22 | **100 / 100** | 0 |
| **Total** | **60.12 / 77.43** | **71.38 / 89.58** | 27 |

`api.ts` is still below the "high" tier because the response interceptor (token-refresh queue, 401 retry, `isPublicPath` redirect) and the `apiClient.create()` config have no test coverage at all — every survivor in that file is in code the existing suite never executes. Killing those needs a real axios-adapter test harness, which is its own PR.

### Hook tests, briefly

- `useWebSocket` (21 cases) — cookie/token URL building, frame parsing for `chat_message` / `notification`, JSON parse-error swallow, send-when-OPEN, reconnect backoff (with the 8s short-open path), cap-at-5 retries, code 4xxx no-retry, stable-timer reset, URL change reopen, unmount cleanup.
- `usePolling` (8) — `isLoading` only on first call vs `isRefreshing` on subsequent, manual refresh, AbortError/CanceledError/ERR_CANCELED swallow, real-error capture, `enabled: false` short-circuit, `visibilitychange` re-run, cleanup aborts and clears the interval.
- `useNotificationSocket` (8) — auth-gated open, `addNotification` + sonner toast on incoming frames, ignores non-notification types and parse errors, no-retry on 4xxx, transient reconnect, disconnect when auth flips to logged-out.
- `useAcquireLocation` (6) — skips when `geoLocation` already set, skips without consent, success path writes to the store, error swallow, unmount race (resolution after unmount must not write).
- `useMyReports` (6) — load, generic error, non-`Error` rejection fallback message, abort on unmount, post-abort silence.

### Mutation-deepening, briefly

- `cookies.ts` — assert default `deleteCookie` path is `/` via a temporary `document.cookie` setter spy (jsdom collapses cookies on read).
- `conversationAPI.ts` — non-empty `participants` / `messages` survive the `?? []` fallback intact; `eventChat.getMessages` forwards the `AbortSignal` exactly.
- `dateTime.ts` — `m=0` collapse in `formatDuration`, plus the boundaries at exactly 24h and 48d in `chatTimestamp`.
- `eventUtils.ts` — AM/PM marker on `formatEventDateTime`, case-insensitive no-AM/PM check on `formatGroupOfferDateTime`, `diff == 0` in `timeUntilEvent`, exactly-60min and exactly-24h boundary kicks, exact-now strict-greater-than in `isEventBanned` / `isOrganizerBanned`, month/day/year asserts in `formatBanExpiry`.
- `api.ts` — Array and string response bodies fall through (the `!Array.isArray` and `typeof === 'object'` guards), plural "messages" key membership in `knownTopKeys`, `", "` vs `". "` join separators in field-error rendering.

Two equivalent mutants:
- `dateTime.ts:44` — `Number.isInteger` short-circuit (the floor/round fallthrough produces the same `${h}h` for integer inputs).
- `eventUtils.ts:30` — `isNearlyFull` guard (max ≤ 0 fallthrough yields NaN/Infinity which fail the `pct < 1.0` check anyway).

### Round 2 test plan additions

- [ ] `frontend-test` job stays green (568 vitest cases, was 502)
- [ ] `frontend-mutation` job — score ≥ 60 (locally 71.38)
- [ ] `frontend-coverage` — thresholds unchanged, all four metrics rose

### DIFF
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -226,3 +226,30 @@ jobs:
       - name: Stop MinIO
         if: always()
         run: docker rm -f ci_minio || true
+
+  assert-sweep:
+    runs-on: ubuntu-latest
+    timeout-minutes: 4
+    name: assert-sweep guardrail
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install libcst
+        run: pip install "libcst>=1.4"
+
+      - name: Run codemod in --check mode
+        working-directory: backend
+        run: python -m scripts.codemods.sweep_status_assertions --check api/tests/integration/
+
+      - name: Forbid TestCase subclasses in unit tests
+        working-directory: backend
+        run: |
+          set -e
+          if grep -rEn 'class \w+\((TestCase|APITestCase|TransactionTestCase|HypothesisTestCase)\)' api/tests/unit/ --include='*.py' 2>/dev/null; then
+            echo "::error::TestCase subclasses found in api/tests/unit/."
+            exit 1
+          fi
--- a/.github/workflows/ci-mutation.yml
+++ b/.github/workflows/ci-mutation.yml
@@ -27,7 +27,16 @@ permissions:
   pull-requests: write
 
 jobs:
+  # Backend mutmut runs on a nightly schedule + manual dispatch only — NOT
+  # on every PR. The CI scope (currently just badge_utils.py to fit the
+  # 45-min job timeout) doesn't earn its ~10 min of runner time per PR
+  # when the rest of the hot-module list isn't actually being gated. Keep
+  # the wiring so the bootstrap stays exercised by the nightly run, and
+  # so anyone debugging mutation testing can trigger it on demand from
+  # the Actions tab. To temporarily run it on a PR, dispatch this
+  # workflow against the PR's branch from the GitHub UI.
   backend-mutmut:
+    if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     timeout-minutes: 45
     services:
--- a/.github/workflows/ci-mutation.yml
+++ b/.github/workflows/ci-mutation.yml
@@ -75,6 +75,36 @@ jobs:
           sudo apt-get update -qq
           sudo apt-get install -y --no-install-recommends gdal-bin libgdal-dev libgeos-dev
 
+      - name: Start MinIO
+        run: |
+          docker run -d --name ci_minio \
+            -p 9000:9000 -p 9001:9001 \
+            -e MINIO_ROOT_USER=minioadmin \
+            -e MINIO_ROOT_PASSWORD=minioadmin123 \
+            --health-cmd "curl -f http://localhost:9000/minio/health/live || exit 1" \
+            --health-interval 5s \
+            --health-timeout 3s \
+            --health-retries 20 \
+            minio/minio:latest server /data --console-address ":9001"
+
+      - name: Wait for MinIO
+        run: |
+          for i in $(seq 1 30); do
+            STATUS=$(docker inspect --format='{{if .State.Health}}{{.State.Health.Status}}{{else}}{{.State.Status}}{{end}}' ci_minio)
+            if [ "$STATUS" = "healthy" ]; then
+              echo "MinIO is healthy."
+              break
+            fi
+            if [ "$STATUS" = "unhealthy" ]; then
+              echo "::error::MinIO became unhealthy"
+              docker logs ci_minio
+              exit 1
+            fi
+            echo "Waiting for MinIO... status=$STATUS ($i/30)"
+            sleep 2
+          done
+
+
       - name: Cache pip
         uses: actions/cache@v4
         with:
@@ -97,7 +127,6 @@ jobs:
         run: |
           .venv/bin/mutmut run || true
           .venv/bin/mutmut results > mutmut-results.txt
-          .venv/bin/mutmut html || true
 
       - name: Upload mutmut report
         if: always()
@@ -106,7 +135,6 @@ jobs:
           name: backend-mutation
           path: |
             backend/mutmut-results.txt
-            backend/html/
 
       # Only post a PR comment if mutmut produced at least one scored
       # mutant. With the current Django app-registry issue, every line in
--- a/.github/workflows/ci-mutation.yml
+++ b/.github/workflows/ci-mutation.yml
@@ -127,6 +127,11 @@ jobs:
         run: |
           .venv/bin/mutmut run || true
           .venv/bin/mutmut results > mutmut-results.txt
+          # `mutmut results` only prints SURVIVED/timeout/suspicious mutants,
+          # not the killed ones — counting it would always read killed=0.
+          # `export-cicd-stats` writes a JSON tally with the actual totals,
+          # which is what the PR comment script reads.
+          .venv/bin/mutmut export-cicd-stats || true
 
       - name: Upload mutmut report
         if: always()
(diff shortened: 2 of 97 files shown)


mzyavuz (approved):
Great work overall. The codemod approach is the right call for a sweep this size, the two flake root causes are correctly identified, and the new tests are really good. They make it easy to see where coverage is still missing.

Approving it!

Issue #538, opened 2026-05-08
Bug: Group offer leaks time credits, provider gets paid once but every participant pays full


yusufizzetmurat:
Group offers don't balance the books. Every participant gets debited the full duration but the provider is only credited once total, so hours just disappear from the system on every group session.

Example: 2-hour group offer, 3 participants.
- Each participant is debited 2h on approval (`provision_timebank`, `backend/api/utils.py:85-117`, called via `ensure_accepted_handshake_reservation`).
- On completion, `complete_timebank_transfer` skips the per-handshake provider credit for group services (`utils.py:311`).
- When the last one completes, the provider gets credited `service.duration` once: 2h, not 6h (`utils.py:365-389`).

So 6h leaves participants, 2h reaches the provider, 4h is gone.

## Expected
Books balance after a group session. Either credit the provider per completed participant, or charge each participant a per-seat share, but the totals must match.

## How to reproduce
1. Provider posts a One-Time group Offer, duration 2h, max_participants 3.
2. Three users express interest, provider initiates details, each approves.
3. All three confirm completion.
4. Sum the `TransactionHistory` deltas across all four users. Currently -4h.

## Acceptance
- [ ] Sum of debits = sum of credits across all users involved in a group offer.
- [ ] Cancelled-mid-flow refunds still match what was debited.
- [ ] One-to-one Offer/Need accounting unchanged.
- [ ] Test covers the 3-participant case end-to-end.

yusufizzetmurat:
Closing as not-a-bug. After internal review, the asymmetric time-credit accounting on group offers is intentional design, not a leak.

The mental model: a yoga teacher giving one 1-hour class to four students earns one hour of credit, not four. Each student pays for their seat (full duration each), but the provider only delivered one session, so they receive one duration credit total. The surplus hours are a deliberate system sink.

This will be locked in with a regression test in the upcoming group-offer correctness PR so a future audit doesn't accidentally flip the behaviour.

PR #543, opened 2026-05-08
fix(offers): group offer correctness pass (closes #520, #521, #539, #540, #541)

### LINKED ISSUE #520: Bug: Group offers disappear from feed after one accepted handshake
Group offers are designed to support multiple participants until the configured capacity is reached. A single accepted handshake should not make the listing unavailable if there are still open slots.

## Current Behavior

When one handshake on a group offer becomes `accepted`, the group offer disappears from the public feed/discovery list.

## Expected Behavior

A group offer should remain visible in the feed while it still has available capacity. It should be hidden only when it is full, expired, deleted, cancelled, or otherwise no longer discoverable.

## Scope

- Review service feed filtering for group offers.
- Check accepted-handshake capacity logic.
- Ensure one-to-one offer hiding rules are not incorrectly applied to group offers.

## Tasks

- Reproduce with a group offer where `max_participants > 1`.
- Accept one participant and verify remaining capacity.
- Fix backend feed/query filtering if accepted status hides the service too early.
- Verify frontend does not apply additional incorrect filtering.

## Acceptance Criteria

- A group offer with remaining slots stays visible after one accepted handshake.
- A full group offer is hidden from public discovery.
- One-to-one offers keep their existing accepted/locked visibility behavior.
- Regression tests cover partially filled and full group offers.

## Test Plan

- Backend: add integration test for partially accepted group offer visibility.
- Backend: add integration test for full group offer hidden from feed.
- Frontend E2E: create group offer, accept one participant, confirm it remains discoverable.

### LINKED ISSUE #521: Bug: Handshake can be moved directly to accepted without proposal approval flow
Handshake acceptance should require the normal session proposal flow: the service owner proposes session details, then the other participant reviews and approves them. This must be enforced by the backend, not only by frontend UI state.

## Current Behavior

From the service detail flow, a handshake can be moved directly to `accepted` without first proposing session details and receiving approval from the other party. This indicates both a frontend guard issue and a backend validation gap.

## Expected Behavior

The backend should reject any direct transition to `accepted` unless the required proposal and approval steps have been completed. The frontend should only expose actions that match the valid backend state machine.

## Scope

- Review handshake status transition endpoints.
- Validate `initiate`, `approve`, and any direct accept/acceptance paths.
- Ensure required session details exist before approval.
- Ensure the approving user is the correct counterparty, not the proposer.

## Tasks

- Identify endpoint(s) that can set `accepted` too early.
- Add backend state-machine validation for proposal-before-approval.
- Harden frontend service detail actions so invalid transitions are not shown.
- Add regression coverage for bypass attempts.

## Acceptance Criteria

- A pending handshake cannot become `accepted` without proposed session details.
- The proposer cannot approve their own proposal.
- Backend rejects invalid direct acceptance even if called manually.
- Frontend no longer exposes a bypass path from service detail.

## Test Plan

- Backend: integration test direct accept/approve without proposal returns 400.
- Backend: integration test proposer cannot approve their own proposal.
- Backend: integration test valid propose -> counterparty approve still succeeds.
- Frontend E2E: verify service detail follows the required proposal/review flow.

### LINKED ISSUE #539: Bug: Group offer can exceed max_participants under concurrent accept/approve
The capacity check happens at express-interest time but not consistently when a handshake actually flips to `accepted`. Two ways the cap gets blown past.

**Path A: `approve()` doesn't check capacity at all.** When the requester approves provider-initiated session details (`backend/api/services.py:605-666`), the handshake transitions straight to `accepted`. There's no `_check_max_participants` call. If the provider accepted other people in the meantime, the offer can end up over capacity.

**Path B: `accept()` validates after writing.** In `accept()` (`services.py:729-801`), `handshake.status = 'accepted'` is saved on line 757, then capacity is counted on line 760. The service row isn't locked, so two simultaneous accepts can both pass the count. Recurrent group offers are worse: the cap recheck only runs in the One-Time branch (line 759), so for `schedule_type='Recurrent'` it never runs at acceptance.

## Expected
Both `approve()` and `accept()` lock the service row and re-check `max_participants` before flipping a handshake to `accepted`. Recurrent offers enforce the cap too.

## How to reproduce (Path A)
- Group offer, max=2. Three users express interest, all with provider-initiated details.
- Provider accepts users B and C through the accept endpoint.
- User A approves their own pending one. A goes to `accepted`. Total accepted: 3.

## Acceptance
- [ ] `approve()` re-checks capacity inside its transaction.
- [ ] `accept()` locks the service and re-checks capacity before saving.
- [ ] Recurrent group offers also enforce the cap on accept.
- [ ] Returns 400 when capacity would be exceeded; no partial state.

### LINKED ISSUE #540: Bug: Group offer max_participants can be lowered below the number already accepted
`validate_max_participants` (`backend/api/serializers.py:745-751`) only checks the value is between 1 and 100. On update it doesn't compare against the current accepted count, so a provider can drop a 5-participant offer down to 2 while five participants are already accepted.

Knock-on effects:
- Capacity-fill UI shows nonsense (negative remaining slots).
- The "service goes to Agreed once full" branch in `accept()` (`services.py:787-788`) misfires.
- The deny-others-on-fill behaviour can fire prematurely on the next acceptance.

## Expected
Lowering `max_participants` below the current capacity-consuming count (`accepted | completed | reported | paused`) is rejected with a clear message.

## How to reproduce
1. Recurrent group Offer, max=3.
2. Three users get accepted.
3. PATCH the service: `{"max_participants": 1}`. Currently returns 200.

## Acceptance
- [ ] Object-level validator rejects the change with a 400 naming the floor.
- [ ] Increasing the cap is unaffected.
- [ ] One-to-one Offer/Need (max=1) is unaffected.
- [ ] EditServiceForm prevents submitting below the floor.

### LINKED ISSUE #541: Bug: Denied requesters on a filled group offer get no notice when the offer is cancelled
When a One-Time group offer fills, `accept()` flips the leftover pending handshakes to `denied` and tells those users their slot is full (`services.py:766-785`). Fine. But if the provider later cancels the whole service, those denied users get nothing: no second notification, no update. The offer just silently vanishes from their world.

`ServiceViewSet.destroy` (`views.py:2725-2750`) only blocks deletion when there are active handshakes (`pending | accepted | checked_in | attended`); `denied` isn't in that set, so cancellation goes through without touching them.

## Expected
On service cancellation, every distinct user with a `denied` handshake for that service gets a notification that the offer was cancelled.

## How to reproduce
1. Group offer, max=1. Two users express interest.
2. Provider accepts user A; user B gets the "all slots filled" notification.
3. Provider deletes the service.
4. User B's notifications: nothing new.

## Acceptance
- [ ] On soft-delete, fan out a notification to every distinct user with a denied handshake for that service.
- [ ] One notification per user even if they had multiple denied handshakes.
- [ ] Existing accepted-handshake cancellation flow unchanged.

yusufizzetmurat:
## Summary
Closes #520
Closes #521
Closes #539  
Closes #540 
Closes #541 

A bundled fix for the group-offer + handshake correctness bugs. They all touch the same handful of files (`services.py`, `serializers.py`, `views.py` plus the matching frontend), so splitting them would just create five PRs that step on each other.

## Changes

**Backend**
- `HandshakeService.accept`: takes a service-row lock, rechecks `max_participants` before flipping status, and now rejects Offer/Need without `provider_initiated`. Direct accept stays as the canonical Event path.
- `HandshakeService.approve`: locks the service, rechecks capacity, and on capacity-fill drives the One-Time → Agreed transition + denies remaining pending (parity with `accept`).
- `_enforce_capacity` / `_close_remaining_when_full`: shared helpers so the two paths stay in sync; capacity rules now apply to Recurrent group offers too.
- `invalidate_service_lists` / `invalidate_user_services` on both transitions so the cached feed/dashboard reflects the new state immediately (#520).
- `ServiceSerializer.validate`: on update, reject `max_participants` below the current count of capacity-consuming handshakes.
- `ServiceViewSet.destroy`: fan a `handshake_cancelled` notification out to every distinct user with a denied handshake, so they hear back when the offer is pulled.

**Frontend**
- `ServiceDetailPage`: the direct Accept button only renders for Event handshakes. Offer/Need providers go through the chat-side Initiate flow (already exists).
- `ServiceForm`: in edit mode, the `max_participants` min is pinned to the current `participant_count` with a helper line.

**Tests**
- New `backend/api/tests/unit/test_group_offer_correctness.py` covers each fix plus a regression lock for the intentional group-offer time-credit asymmetry (#538 closed wontfix — receivers pay full, provider earns once, surplus is a system sink).
- Existing capacity tests switch from direct `accept()` to the propose→approve helper.

## How to Test
1. `make test-unit` — `test_group_offer_correctness.py` plus the updated capacity tests pass.
2. Manually:
   - Provider creates a One-Time group Offer, max=4, 2h.
   - Two requesters express interest, provider goes through Initiate in chat for each, both approve.
   - Service still appears in a logged-out viewer's feed (#520).
   - Direct POST to `/handshakes/{id}/accept/` for an Offer returns 400 (#521).
   - Provider edits max from 4 to 1 → 400 with floor message (#540).
   - Provider deletes the service after one accept + two denied → both denied users see a `handshake_cancelled` notification (#541).
3. Concurrency: two simultaneous approves on the last open seat → exactly one succeeds, the other gets `INVALID_STATE` for capacity (#539).

## Checklist
- [x] Tests added / updated
- [ ] Documentation updated (no doc surface affected)
- [x] No breaking changes to Event accept or one-to-one Offer/Need happy paths


mzyavuz (approved):
Tested all acceptance criteria manually, everything works as expected:

- Direct `POST /handshakes/{id}/accept/` on an Offer returns 400; Event accept still works through the direct path.
- Two concurrent approves on the last open slot: exactly one succeeds, the other gets rejected for capacity.
- Lowering `max_participants` below the current accepted count returns 400 with the floor message.
- Deleting a service after a mix of accepted + denied handshakes correctly fans out `handshake_cancelled` notifications to denied users.
- After all slots are filled and the offer is cancelled, a new user can still express interest and go through the flow — no stale state blocking them.

**Approved.**

PR #571, opened 2026-05-10
fix: settle group-offer credit on cancellation and dedupe location form

### LINKED ISSUE #506: Bug: Group offer form has duplicate location inputs
Group offer form shows two location inputs, which confuses users.

## Expected
Show only one clear location flow for group offers.

## Tasks
- Remove/hide the duplicate location input.
- Keep in-person address/map selection working.
- Hide physical location fields for online offers.

## Acceptance
- No duplicate location input appears.
- Group offer create/edit still works.
- Location data submits correctly.

### LINKED ISSUE #557: Bug: Group offer tx issue
Group offer tx's credit transfer only transffered if first user completed the offer. It's not best way to complete time transfer. Need fix.

yusufizzetmurat:
This PR closes two related group-offer correctness bugs.

## Bug A — Group offer credit transfer fires only on first completion (#557)

**Real-bug description.** Group offers follow a documented asymmetric model: receivers pay per seat (escrowed when each handshake is approved) and the provider earns ``service.duration`` exactly once after every active handshake reaches a terminal state. The surplus is a deliberate system sink. The completion path correctly waited for every active handshake to settle, but it relied solely on cancellations *not* happening — when a trailing handshake was cancelled after earlier participants completed, no recheck ran and the provider's payout was never triggered. End result: a service that was actually delivered to multiple receivers could leave the provider unpaid.

**Root cause.** ``cancel_timebank_transfer`` only refunded the cancelling receiver and flipped the handshake to ``cancelled``. It did not look at the rest of the service to see whether this cancellation just drained the last active slot on a one-time group offer where receivers had already completed.

**Fix.** Extracted the single-payout settlement into ``_settle_group_offer_provider_payout`` (idempotent — refuses to pay if a transfer transaction already exists for the service, and refuses to pay when no completed handshake exists). Both ``complete_timebank_transfer`` and ``cancel_timebank_transfer`` now call it after the active-count check, so the payout fires from whichever transition drains the last active slot.

**Verification.**
- Unit regression: ``test_group_one_time_offer_pays_provider_when_trailing_handshake_cancels`` (3 receivers, 2 complete, 3rd cancels → provider paid once, receivers not refunded — system sink intact).
- Unit regression: ``test_group_one_time_offer_skips_payout_when_no_one_completed`` (everyone cancels → provider not paid, no transfer history).
- API regression: ``test_group_offer_settles_provider_after_each_receiver_completes`` exercises the full ``/api/handshakes/<id>/confirm/`` flow across two participants.
- Existing ``test_group_one_time_offer_transfers_only_once_and_all_receivers_pay`` still passes (no behaviour change on the all-completed path).

Closes #557

## Bug B — Group offer form has duplicate location inputs (#506)

**Real-bug description.** The fixed in-person group offer form rendered two separate location pickers — a "Public district / area" picker and a "Exact address for session details" picker — and required users to fill in both. This was confusing on its own and meant any change to one had to be mirrored in the other.

**Root cause.** ``ServiceForm.tsx`` rendered both ``LocationSearch`` controls inside the ``locType === 'In-Person' && isFixedGroupOffer`` branch. The two values are not actually independent: the public district can be derived from the exact address (via the existing ``syncPublicFromExact`` helper).

**Fix.** Collapsed the in-person fixed group offer flow to a single canonical "Meeting address" picker with the existing search + map, and let ``syncPublicFromExact`` derive ``location_area`` / ``location_lat`` / ``location_lng`` from the exact address. Online fixed group offers continue to render only the meeting-link input — no physical-location helpers leak in.

**Verification.**
- ``frontend/src/test/components/ServiceForm.location.test.tsx`` covers all three relevant configurations: regular Offer renders a single Address picker, fixed in-person group offer renders a single "Meeting address" picker plus map (and zero remaining duplicates), online fixed group offer renders only the meeting-link input with no map or address picker.
- ``cd frontend && npm run lint && npx tsc --noEmit -p tsconfig.app.json`` is clean.

Closes #506

## Test plan

- [x] ``cd backend && python -m pytest api/tests/unit/test_utils.py api/tests/unit/test_capacity_and_agreed.py api/tests/integration/test_handshake_api.py`` — 111 passed
- [x] ``cd frontend && npx vitest run src/test/components`` — 123 passed
- [x] ``cd frontend && npm run lint`` — clean
- [x] ``cd frontend && npx tsc --noEmit -p tsconfig.app.json`` — clean

### DIFF
--- a/backend/api/tests/integration/test_calendar_api.py
+++ b/backend/api/tests/integration/test_calendar_api.py
@@ -220,7 +220,14 @@ class TestMeCalendarItems:
         active_requester = UserFactory()
         svc = ServiceFactory(user=provider, type='Offer', duration=Decimal('2.00'), max_participants=3)
         scheduled = timezone.now() + timedelta(days=5)
-        completed_at = timezone.now() - timedelta(days=10)
+        # Pin the completion time to mid-day. The calendar models a completed
+        # session as ending at completed_at and starting duration hours earlier;
+        # if completed_at falls in the small hours the computed start crosses
+        # midnight and lands on the previous day, breaking the same-day
+        # assertion below.
+        completed_at = (timezone.now() - timedelta(days=10)).replace(
+            hour=12, minute=0, second=0, microsecond=0
+        )
         handshake = HandshakeFactory(
             service=svc, requester=requester, status='completed',
             scheduled_time=scheduled,
--- a/backend/api/tests/integration/test_handshake_api.py
+++ b/backend/api/tests/integration/test_handshake_api.py
@@ -457,15 +457,14 @@ class TestHandshakeViewSet:
         provider.refresh_from_db()
         assert provider.timebank_balance > Decimal('5.00')
 
-    def test_group_offer_settles_provider_after_all_receivers_complete(self):
+    def test_group_offer_settles_provider_on_first_completion(self):
         """End-to-end coverage for the asymmetric group-offer settlement.
 
-        The provider earns ``service.duration`` exactly once. Each receiver
-        pays per seat (escrowed at acceptance) and is not refunded when their
-        own handshake completes — the surplus is the documented system sink.
-        Regression for #557: confirms the payout fires reliably only after
-        all participant handshakes complete, not while other participants are
-        still in-flight.
+        Pays the provider on the FIRST completion, not the last (sgunes
+        review). Settlement is idempotent: completing handshake2 must not
+        produce a second transfer. Each receiver pays per seat (escrowed at
+        acceptance) and is not refunded when their own handshake completes —
+        the surplus is the documented system sink.
         """
         provider = UserFactory(timebank_balance=Decimal('0.00'))
         receiver1 = UserFactory(timebank_balance=Decimal('5.00'))
@@ -500,7 +499,7 @@ class TestHandshakeViewSet:
 
         client = AuthenticatedAPIClient()
 
-        # Both sides confirm handshake1 — provider must NOT be paid yet.
+        # Both sides confirm handshake1 — provider settles on this completion.
         client.authenticate_user(provider)
         assert_api_response(client.post(f'/api/handshakes/{handshake1.id}/confirm/'), 200)
         client.authenticate_user(receiver1)
@@ -509,11 +508,17 @@ class TestHandshakeViewSet:
         handshake1.refresh_from_db()
         provider.refresh_from_db()
         assert handshake1.status == 'completed'
-        assert provider.timebank_balance == Decimal('0.00'), (
-            'Group offer must wait until every participant settles before paying the provider'
+        assert provider.timebank_balance == Decimal('3.00'), (
+            'Group offer must pay the provider on the first completion, not the last'
         )
+        # The single transfer row already exists by this point.
+        assert TransactionHistory.objects.filter(
+            user=provider,
+            transaction_type='transfer',
+            handshake__service=service,
+        ).count() == 1
 
-        # Both sides confirm handshake2 — payout fires on the last completion.
+        # Both sides confirm handshake2 — settlement must be idempotent.
         client.authenticate_user(provider)
         assert_api_response(client.post(f'/api/handshakes/{handshake2.id}/confirm/'), 200)
         client.authenticate_user(receiver2)
--- a/backend/api/tests/integration/test_handshake_api.py
+++ b/backend/api/tests/integration/test_handshake_api.py
@@ -457,15 +457,15 @@ class TestHandshakeViewSet:
         provider.refresh_from_db()
         assert provider.timebank_balance > Decimal('5.00')
 
-    def test_group_offer_settles_provider_after_each_receiver_completes(self):
+    def test_group_offer_settles_provider_after_all_receivers_complete(self):
         """End-to-end coverage for the asymmetric group-offer settlement.
 
         The provider earns ``service.duration`` exactly once. Each receiver
         pays per seat (escrowed at acceptance) and is not refunded when their
         own handshake completes — the surplus is the documented system sink.
-        Regression for #557: confirms the payout fires reliably across
-        per-handshake completion, even when other participants are still
-        in-flight.
+        Regression for #557: confirms the payout fires reliably only after
+        all participant handshakes complete, not while other participants are
+        still in-flight.
         """
         provider = UserFactory(timebank_balance=Decimal('0.00'))
         receiver1 = UserFactory(timebank_balance=Decimal('5.00'))
--- a/backend/api/tests/integration/test_handshake_api.py
+++ b/backend/api/tests/integration/test_handshake_api.py
@@ -436,26 +436,104 @@ class TestHandshakeViewSet:
             provider_initiated=True,
             requester_initiated=True
         )
-        
+
         client = AuthenticatedAPIClient()
         client.authenticate_user(provider)
-        
+
         response = client.post(f'/api/handshakes/{handshake.id}/confirm/')
         assert_api_response(response, 200)
-        
+
         handshake.refresh_from_db()
         assert handshake.provider_confirmed_complete is True
-        
+
         client.authenticate_user(requester)
         response = client.post(f'/api/handshakes/{handshake.id}/confirm/')
         assert_api_response(response, 200)
-        
+
         handshake.refresh_from_db()
         assert handshake.status == 'completed'
         assert handshake.receiver_confirmed_complete is True
-        
+
         provider.refresh_from_db()
         assert provider.timebank_balance > Decimal('5.00')
+
+    def test_group_offer_settles_provider_after_each_receiver_completes(self):
+        """End-to-end coverage for the asymmetric group-offer settlement.
+
+        The provider earns ``service.duration`` exactly once. Each receiver
+        pays per seat (escrowed at acceptance) and is not refunded when their
+        own handshake completes — the surplus is the documented system sink.
+        Regression for #557: confirms the payout fires reliably across
+        per-handshake completion, even when other participants are still
+        in-flight.
+        """
+        provider = UserFactory(timebank_balance=Decimal('0.00'))
+        receiver1 = UserFactory(timebank_balance=Decimal('5.00'))
+        receiver2 = UserFactory(timebank_balance=Decimal('5.00'))
+        service = ServiceFactory(
+            user=provider,
+            type='Offer',
+            duration=Decimal('3.00'),
+            schedule_type='One-Time',
+            max_participants=2,
+        )
+        handshake1 = HandshakeFactory(
+            service=service, requester=receiver1, status='accepted',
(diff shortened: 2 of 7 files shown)


sgunes16:
still wait first completion. bug is alive

sgunes16:
506 issue is fixed btw

sgunes16 (changes requested):
# Group Offer Settlement Bug — Root Cause & Simple Fix

## What the user is hitting

1. Provider creates a group offer (`max_participants > 1`, `schedule_type = One-Time`, `duration = 2h`).
2. Zeynep applies first → handshake accepted, `-2h` escrowed from Zeynep.
3. Cem applies second → handshake accepted, `-2h` escrowed from Cem.
4. Cem confirms completion (both sides) **before** Zeynep does anything.

**Expected.** The provider's `2h` payout fires on Cem's completion. A `transfer` row appears in the provider's transaction list. Cem's session leaves "active" and shows up on the completed-exchange side. Zeynep's later complete-or-cancel is independent — it does **not** affect what the provider already received.

**Actual.**
- Provider's balance unchanged.
- No `transfer` row anywhere on the provider side. From the provider's seat, Cem's session is invisible until Zeynep also settles.
- Cem's own `provision` row does correctly flip to "Time used / completed exchange" once `handshake.status === 'completed'` (this is rendered live via `isCompletedTransactionContext`), but from the **provider's** seat there is no acknowledgement.

## The intended model (which we are keeping)

`backend/api/utils.py` documents an **asymmetric group-offer model** and that part is correct:

- Each receiver pays one full seat (`provisioned_hours = service.duration`) at acceptance.
- The provider is paid **exactly once** for the whole group session — a single `service.duration` credit.
- The surplus from the extra seats is a deliberate **system sink**.

We want to keep all three properties. The bug is **not** in the asymmetric model. It is in **when** the single payout fires.

## The actual root cause

`complete_timebank_transfer` only fires the group payout once *every* active handshake on the service has reached a terminal state:

```python
if _is_group_one_time_service(service) and active_count_after == 0:
    _settle_group_offer_provider_payout(service, handshake)
```

So Cem completing while Zeynep is still `accepted` leaves `active_count_after == 1`, and the payout never fires.

But `_settle_group_offer_provider_payout` is already **idempotent**:

```python
already_paid = TransactionHistory.objects.filter(
    user=provider, transaction_type='transfer', handshake__service=service,
).exists()
if already_paid:
    return False

completed_exists = Handshake.objects.filter(
    service=service, status='completed',
).exists()
if not completed_exists:
    return False
```

It refuses to pay twice for the same service, and refuses to pay before anyone has completed. That is exactly the contract we need: **fire on every completion; the helper itself decides whether to act**.

The `active_count_after == 0` gate adds nothing — it just delays a payout the helper would already gate correctly.

PR #571's `fc685e5` partially addressed this from the cancellation side (a trailing cancellation on a service where someone already completed will now drain the active count and trigger the helper). It does **not** help the case the user is hitting, where the trailing handshake is still in flight (`accepted`), not cancelled.

The PR comment "still wait first completion. bug is alive" is correct.

### Why "history tarafına geçmiyor" on the provider side

For the provider, no `transfer` row exists for Cem yet (the helper didn't fire). The active-list filter is:

```ts
const ACTIVE_HANDSHAKE_STATUSES = new Set(['accepted', 'checked_in', 'attended'])
```

Cem's `completed` handshake is filtered out of "active". And there's no `transfer` row to put on the completed side. From the provider's seat, Cem's session is in **neither** section — unrendered until Zeynep also settles.

For Cem himself the UI is fine: his `provision` row reads `handshake_status === 'completed'` live and renders as "Time used / completed exchange".

## The Simple Fix

Drop the `active_count_after == 0` gate from `complete_timebank_transfer`. Let the helper decide:

```python
# BEFORE
if _is_group_one_time_service(service) and active_count_after == 0:
    _settle_group_offer_provider_payout(service, handshake)

# AFTER
if _is_group_one_time_service(service):
    _settle_group_offer_provider_payout(service, handshake)
```

That is the entire backend change. The asymmetric model and the system sink stay exactly as they are:

- The helper still pays the provider **once** per service (idempotent guard).
- The helper still refuses to pay if **no one has completed** (cancellation-only path keeps the provider unpaid).
- Receivers are still **not refunded** on their own completion — surplus seats remain a system sink.

The trailing-cancellation hook in `cancel_timebank_transfer` becomes redundant (the helper would already have fired on whichever completion came first), but it is also harmless because of the idempotency guard. Either keep it as defence-in-depth or remove it for cleanliness — it does not change observable behaviour.

## Behavioural Diff

Group offer with `duration = 2h`, `max_participants = 3`, `Zeynep`/`Cem`/`Ayşe` each escrow `2h` at acceptance.

| Scenario | Today | After the fix |
| --- | --- | --- |
| Cem confirms completion while Zeynep is still `accepted` | Provider receives `0h`. Cem's session is invisible on the provider side. | Provider receives `+2h` immediately. `transfer` row appears with Cem as counterpart. |
| Zeynep then completes | Provider finally receives `+2h` (single asymmetric payout fires now). | No extra credit (idempotent). Zeynep's escrow is gone — surplus to system sink. |
| Zeynep cancels instead of completing | Trailing-cancel path fires the single payout: provider gets `+2h`. | Provider already has the `+2h` from Cem. Cancellation only refunds Zeynep. |
| Nobody completes — all three cancel | Provider receives `0h` (helper sees no completed handshake). | Identical: helper still sees no completion, refuses to pay. |
| All three complete | Provider receives `+2h` once. | Same: `+2h` once on the first completion; subsequent completions are no-ops. |

End state for the provider is **always `+2h` (or `0h` if nobody completed)** — the system sink is fully preserved. Only the *timing* changes: the payout now lands on the first completion instead of waiting for everyone to finish.

## Files to Touch

### Backend

- `backend/api/utils.py`
  - In `complete_timebank_transfer`, remove `and active_count_after == 0` from the group-payout call. (One-line change.)
  - Optional cleanup: drop the trailing-settle block at the end of `cancel_timebank_transfer` since the helper now always fires on the first completion. Keep it if you prefer defence-in-depth — the idempotency guard makes it a no-op in normal flow.

### Backend tests (existing tests encode "wait for everyone")

- `backend/api/tests/unit/test_utils.py`
  - `test_group_one_time_offer_transfers_only_once_and_all_receivers_pay` — final balance is unchanged (`+3h` once), but the assertion **after the first completion** flips from "provider not paid yet" to "provider paid `+3h`". After the second completion, balance must remain `+3h` (idempotent).
  - `test_group_one_time_offer_pays_provider_when_trailing_handshake_cancels` — flip the timing: payout fires on the first completion, the trailing cancellation only refunds the canceller. Final state is identical (`+3h` to provider, last receiver refunded).
  - `test_group_one_time_offer_skips_payout_when_no_one_completed` — unchanged in expectation; still asserts no transfer when every handshake cancels.
- `backend/api/tests/integration/test_handshake_api.py`
  - `test_group_offer_settles_provider_after_each_receiver_completes` (current name is misleading — it actually asserts the *delayed* single payout). Update so that after both sides confirm Cem's handshake the provider is already `+3h`, and confirming Zeynep's handshake afterward leaves the balance at `+3h`.

### Frontend

#### Show settled participants as "No change" inside the still-active group session

Today, `fetchActiveAgreements` filters handshakes through

```ts
const ACTIVE_HANDSHAKE_STATUSES = new Set(['accepted', 'checked_in', 'attended'])
```

so a participant whose handshake flips to `completed` **disappears from the active card entirely**. With per-completion timing, this looks wrong from the provider's seat: as soon as Cem confirms, the group card collapses to "1 member · Zeynep" with an `expected_delta` of `+2h`, even though the provider was already paid in full and Zeynep's eventual settlement will not move the balance again.

Required UX:

- Keep `completed` group-offer handshakes inside the participant list of an **otherwise still-active** group card (i.e. the service has at least one sibling handshake in `accepted`/`checked_in`/`attended`).
- Render those settled participants with a delta of `0` and the label **"No change"** in place of the hours pill. The status chip still reads "Completed".
- The grouped header (`"2 members"`) keeps counting the settled participants — they belong to the same session, they are just done.
- Inside `MultiUseDetailsModal`, the same row shows "Completed" for the status, "No change" for the value, and the avatar/name unchanged.

Concretely:

- Expand the filter in `fetchActiveAgreements` so that a `completed` handshake is retained when the service is a one-time group offer **and** at least one sibling handshake is still in an active status (or just unconditionally retained for group offers — the grouping logic handles the rest).
- In `toExpectedAgreement`, branch on `handshake.status === 'completed'`:
  - `expected_delta = 0`
  - `reserved_delta = 0`
  - `note = 'No further change — already settled'`
- In the active-row renderer, when `displayDelta === 0` and `agreement.status === 'completed'`, swap the existing "No hours" copy for "No change". The pill colour stays neutral (`GRAY100` / `GRAY700`).
- The `MultiUseDetailsModal` items already mirror `expected_delta`/`reserved_delta`, so they pick this up for free; only the empty-pill copy needs the same swap.

#### Transaction-list grouping

There is still **at most one** `transfer` transaction per service, so `groupTransactionRows` and `isGroupOfferTransfer` keep their current behaviour. The only timing change the user will notice is that the provider-side `transfer` row appears as soon as the first receiver completes, instead of after the last one settles.

## TL;DR

The asymmetric "provider gets `service.duration` once, surplus is system sink" model is fine. The bug is the **timing gate**: today the payout waits for every active handshake to terminate, even though the helper that issues it is already idempotent. Removing `and active_count_after == 0` in `complete_timebank_transfer` fires the payout on the first completion, leaves the rest of the model untouched, and resolves the user-visible "Cem completed but nothing happens until Zeynep does something" bug.

yusufizzetmurat (commented):
Comments addressed in e9a626e, c8edf11:

- e9a626e fix(api): use service-id idempotency on group-offer payout — switches the `already_paid` guard in `_settle_group_offer_provider_payout` from `handshake__service=service` to `service=service`. `TransactionHistory.handshake` is SET_NULL, so a later handshake deletion no longer hides the prior payout row from the idempotency check; the only remaining failure mode (service deletion) leaves no caller to re-enter the branch.
- c8edf11 docs(api): clarify group-offer payout description — TransactionHistory description now reads "after all participants settled" instead of "completed", honestly covering the trailing-cancellation path the helper also handles.
- Test rename was already shipped in 95bff3a (`test_group_offer_settles_provider_after_each_receiver_completes` → `..._after_all_receivers_complete`), matching the asymmetric "provider earns once on last settlement" design.

Ran `./.venv/bin/pytest api/tests/integration/test_handshake_api.py api/tests/unit/test_utils.py --no-cov` — 81 passed, including the four group-offer regression tests.

sgunes16:
resim
Cem demir should be not at active agreements and no change should be at 

resim
For example above gorup change should be no change

resim
there is some ui bugs also. But we can fix that another pr with mobile


I approved

sgunes16 (approved):

sgunes16:
#610 issue created

Issue #16, opened 2026-02-22
Combine SRS Sections into One Document


yaseminsirin:
All team members have completed their assigned SRS sections. These parts need to be reviewed, aligned in format, and merged into a single consolidated SRS document. The final document should ensure consistency in terminology, structure, and numbering across all sections.

yaseminsirin:
https://github.com/SWE-574/SWE-574-3/wiki/Software-Requirements-Specification

PR #572, opened 2026-05-10
fix(security): cookie sanitization, flag pins, and Report FK audit (#475)

### LINKED ISSUE #475: Construction of a cookie using user-supplied input #16
Constructing cookies from user input can allow an attacker to control a user's cookie. This may lead to a session fixation attack. Additionally, client code may not expect a cookie to contain attacker-controlled data, and fail to sanitize it for common vulnerabilities such as Cross Site Scripting (XSS). An attacker manipulating the raw cookie header may additionally be able to set cookie attributes such as HttpOnly to insecure values.

yusufizzetmurat:
Three concerns under one backend-safety theme: the CodeQL ``py/cookie-injection`` alert (#475), a defensive sweep of the Django cookie security flags, and a preventive audit of the remaining ``Report`` foreign keys after PR #570 covered ``reported_forum_topic``.

## 1. ``CustomTokenRefreshView`` cookie sanitization

**Real bug.** CodeQL alert #16 (``py/cookie-injection``, severity ``high``) flagged ``backend/api/views.py:132`` — ``response.set_cookie('refresh_token', refresh_token, ...)`` was reached with a ``refresh_token`` argument that ultimately came from the request payload.

**Root cause.** ``CustomTokenRefreshView.post`` reads the refresh token with

```python
refresh_token_val = request.COOKIES.get('refresh_token') or request.data.get('refresh')
```

and then, after a successful ``serializer.is_valid()``, fell back to that user-supplied input when SimpleJWT did not rotate:

```python
new_refresh = validated.get('refresh', refresh_token_val)
```

In ``DEBUG`` the project ships with ``ROTATE_REFRESH_TOKENS=False``, so the fallback always fired and the cookie value flowed straight from ``request.data`` / ``request.COOKIES`` into ``Set-Cookie``. Even with rotation on, the fallback path remained reachable on any future config drift.

**Fix.** Re-serialise the validated token through ``str(RefreshToken(refresh_token_val))`` so the cookie value is always constructed from the server-side token object. As a last-line guard, ``_set_auth_cookies`` now refuses any value that does not match the JWT grammar (three base64url segments separated by ``.``); the new ``_is_jwt_shape`` helper documents what counts as canonical and what is rejected. This breaks the source–sink chain CodeQL flagged regardless of any caller's mistake upstream.

**Verification.**

* ``api/tests/security/test_cookie_safety.py::TestCookieSafetyOnRefresh::test_refresh_cookie_does_not_echo_attacker_bytes`` — POSTs a malicious ``refresh`` body field (``garbage; Path=/admin; HttpOnly=false``) with a valid cookie, asserts the response cookie is a fresh JWT and contains none of the injection bytes.
* ``::test_refresh_with_invalid_token_returns_401_and_writes_no_cookie`` — POSTing ``totally-bogus-not-a-jwt`` returns 401 and never lands in any auth cookie.
* The ``TestIsJwtShape`` and ``TestSetAuthCookies`` units cover ``\\n``, ``;``, attribute-injection patterns, and the shape-only contract.
* Existing ``api/tests/integration/test_auth_api.py`` (15 tests) still passes against the new cookie helper.

Closes #475.

## 2. Cookie security flag pins

**Real bug.** The ``HttpOnly`` and ``SameSite`` flags on Django's session and CSRF cookies relied on framework defaults; only the production branch toggled ``SECURE``. A future Django upgrade (or a change in dev-vs-prod branching) could silently downgrade either flag without anyone noticing.

**Root cause.** Defaults are not pins. ``SESSION_COOKIE_HTTPONLY``, ``CSRF_COOKIE_HTTPONLY``, ``SESSION_COOKIE_SAMESITE``, and ``CSRF_COOKIE_SAMESITE`` were never explicit in ``backend/hive_project/settings.py``.

**Fix.** Pin the safe baseline before the production-only block:

```python
SESSION_COOKIE_HTTPONLY = True
CSRF_COOKIE_HTTPONLY = True
SESSION_COOKIE_SAMESITE = 'Lax'
CSRF_COOKIE_SAMESITE = 'Lax'
SESSION_COOKIE_SECURE = False
CSRF_COOKIE_SECURE = False
```

The production block continues to flip ``*_SECURE`` to ``True``. ``Lax`` matches the auth-cookie helper in ``api/views.py`` and is compatible with the SPA's top-level navigations.

**Verification.**

* ``test_cookie_safety.py::TestSetAuthCookies::test_writes_both_cookies_when_inputs_are_well_formed`` confirms the auth helper emits ``HttpOnly`` and a valid ``SameSite``.
* ``::TestCookieSafetyOnRefresh::test_refresh_cookie_flags_are_set_in_dev`` and ``::test_refresh_cookie_flags_include_secure_in_production`` (under ``override_settings(IS_PRODUCTION=True)``) lock the dev / prod posture into the suite.
* ``::TestCookieSafetyOnLogin::test_login_cookies_have_security_flags`` covers the login ``Set-Cookie`` site for the same regression class.

## 3. ``Report`` FK cascade audit (preventive)

**Real bug?** No active bug; documenting the post-#570 contract before a future migration regresses it.

**Root cause / current state.** Each ``Report`` parent and its delete path:

| FK | ``on_delete`` | Parent removal in code | Verdict |
|---|---|---|---|
| ``reporter`` (User) | CASCADE | No public delete; admin only flips ``is_active=False`` | Safe today |
| ``reported_user`` (User) | CASCADE | Same — admin ban is soft | Safe today |
| ``reported_service`` (Service) | CASCADE | ``ServiceViewSet.destroy`` flips ``status='Cancelled'`` | Safe today |
| ``reported_forum_topic`` (ForumTopic) | CASCADE | PR #570 territory (soft-delete ``is_deleted=True``) | Covered by #570 |
| ``reported_forum_post`` (ForumPost) | CASCADE | ``ForumPostViewSet.destroy`` flips ``is_deleted=True`` | Safe today |
| ``related_handshake`` (Handshake) | CASCADE | ``HandshakeViewSet`` exposes no destroy verb | Safe today |

Every report parent except ``reported_forum_topic`` already preserves report rows because the parent row itself never gets hard-deleted via the API. ``reported_forum_topic`` is intentionally left for PR #570 to land first.

**Fix.** No production code change here — the audit is captured in tests so a future regression is caught:

* ``backend/api/tests/integration/test_report_cascade_audit.py::TestReportCascadeAudit`` exercises the public delete / ban path for each parent and asserts the report row survives.
* ``::TestReportCascadeAtModelLevel`` documents the worst case (``Model.objects.delete()`` at the ORM level) so a future ad-hoc data-purge or GDPR migration cannot quietly wipe moderation history without an updated test calling it out.

**Verification.**

```
cd backend && DEBUG=True .venv/bin/python -m pytest --no-cov \\
  api/tests/security/ \\
  api/tests/integration/test_report_cascade_audit.py
# 31 passed
```

## Files changed

* ``backend/api/views.py`` — ``_is_jwt_shape`` helper; ``_set_auth_cookies`` guard; refresh-view re-serialisation.
* ``backend/hive_project/settings.py`` — pinned cookie flags.
* ``backend/api/tests/security/__init__.py`` (new), ``backend/api/tests/security/test_cookie_safety.py`` (new) — 22 cases.
* ``backend/api/tests/integration/test_report_cascade_audit.py`` (new) — 9 cases.

## Test plan

- [x] ``api/tests/security/`` (22 passed)
- [x] ``api/tests/integration/test_report_cascade_audit.py`` (9 passed)
- [x] ``api/tests/integration/test_auth_api.py`` regression (15 passed)
- [ ] CodeQL re-scan closes alert #16 on next push to ``dev``

### DIFF
--- a/backend/api/tests/integration/test_report_cascade_audit.py
+++ b/backend/api/tests/integration/test_report_cascade_audit.py
@@ -27,6 +27,7 @@ from api.tests.helpers.factories import (
     ServiceFactory,
     UserFactory,
 )
+from api.tests.helpers.assertions import assert_api_response, assert_problem_detail
 from api.tests.helpers.test_client import AuthenticatedAPIClient
 
 
@@ -60,7 +61,7 @@ class TestReportCascadeAudit:
 
         client = AuthenticatedAPIClient().authenticate_user(service_owner)
         response = client.delete(f"/api/services/{service.id}/")
-        assert response.status_code == 204, response.content
+        assert_api_response(response, 204)
 
         # Service row survives (soft-delete).
         service.refresh_from_db()
@@ -79,7 +80,7 @@ class TestReportCascadeAudit:
 
         client = AuthenticatedAPIClient().authenticate_user(post.author)
         response = client.delete(f"/api/forum/posts/{post.id}/")
-        assert response.status_code == 204, response.content
+        assert_api_response(response, 204)
 
         post.refresh_from_db()
         assert post.is_deleted is True
@@ -103,11 +104,8 @@ class TestReportCascadeAudit:
         client = AuthenticatedAPIClient().authenticate_user(reporter)
         response = client.delete(f"/api/handshakes/{handshake.id}/")
         # DRF returns 405 (Method Not Allowed) for unsupported HTTP verbs.
-        assert response.status_code in (
-            405,
-            403,
-            404,
-        ), response.content
+        # 403 / 404 are equally valid "no destructive operation here" replies.
+        assert response.status_code in (405, 403, 404)
 
         handshake.refresh_from_db()
         report.refresh_from_db()
@@ -124,7 +122,7 @@ class TestReportCascadeAudit:
 
         client = AuthenticatedAPIClient().authenticate_user(admin)
         response = client.post(f"/api/admin/users/{target.id}/ban/")
-        assert response.status_code == 200, response.content
+        assert_api_response(response, 200)
 
         target.refresh_from_db()
         assert target.is_active is False
@@ -150,12 +148,12 @@ class TestReportCascadeAudit:
         # Self-delete on /api/users/me/ must not be available; 401/403/404/405
         # all encode "no destructive operation here".
         me_delete = client.delete("/api/users/me/")
-        assert me_delete.status_code in (401, 403, 404, 405), me_delete.content
+        assert me_delete.status_code in (401, 403, 404, 405)
 
         # Admin ban on the reporter likewise leaves the row in place.
         admin_client = AuthenticatedAPIClient().authenticate_user(admin)
         ban = admin_client.post(f"/api/admin/users/{reporter.id}/ban/")
-        assert ban.status_code == 200, ban.content
+        assert_api_response(ban, 200)
         reporter.refresh_from_db()
         assert reporter.is_active is False
 
--- a/backend/api/tests/integration/test_report_cascade_audit.py
+++ b/backend/api/tests/integration/test_report_cascade_audit.py
@@ -116,9 +116,7 @@ class TestReportCascadeAudit:
         # ``http_method_names``. Anything else (204 success, 403, 404)
         # would mean DELETE is reachable for a participant and the
         # CASCADE on ``Report.related_handshake`` is exposed.
-        assert response.status_code == 405, (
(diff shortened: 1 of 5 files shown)


sgunes16:
works very well. implementation is clean

sgunes16 (approved):

PR #574, opened 2026-05-10
fix(mobile): auth lifecycle, listener hygiene, and add-offer UX

### LINKED ISSUE #452: Mobile authentication lifecycle is brittle
The mobile API client has no concept of token expiry. Once the JWT goes stale the app gets trapped in a "logged in but everything 401s" state and the user has no way out short of killing and relaunching the app.

**No 401/403 handler in the mobile API client**  `mobile-client/src/api/client.ts:103-114`
`apiRequest` throws a generic `Error(message)` for every non-OK status. No token refresh, no logout dispatch, no kick to the auth stack. Once the access token expires, every subsequent screen shows a generic toast/error and the user is trapped.

**Repro:** Reduce JWT lifetime to 60s in dev, log into the mobile app, leave it idle for two minutes, then tap any screen.

### Suggested fix
Inside `apiRequest`, intercept 401 — try the refresh-token flow once, retry the original request, and on permanent failure clear stored tokens and reset navigation to the auth stack. The backend already exposes a refresh endpoint, this is mostly client glue. Pair this with Issue 1's logging convention so refresh failures show up in our error tracker instead of vanishing

### LINKED ISSUE #453: Mobile UX gaps: push listeners leak, map fallback is silent
Two unrelated mobile bugs that both fall in the platform lifecycle bucket. Worth bundling so we make one focused mobile polish PR.

**a) Push notification listener doesn't tear down old handlers** — `mobile-client/src/hooks/usePushNotifications.ts:139-167`
The effect attaches a `Notifications.addNotificationResponseReceivedListener(...)` keyed on `navigationRef`. When `navigationRef` changes (which happens during cold-start initialization), the cleanup path doesn't reliably remove the previous listener, so we end up with multiple handlers bound. Tapping a single push then fires multiple navigations.

**Repro:** Cold-start the app, sign in, send yourself a push, tap it. If the navigation animation runs twice or the back-stack is suddenly two screens deep, that's this.

**b) MapScreen silently falls back to Istanbul on permission denial** — `mobile-client/src/presentation/screens/MapScreen.tsx:159-181`
On `requestForegroundPermissionsAsync` denial the screen quietly defaults to a hard-coded Istanbul center and stays interactive as if the location were the user's. There's a commented-out block right next to it that looks like an explicit "permission denied" banner that was started and never finished. Outside Istanbul this is going to feel broken — a user in Ankara opens the map and sees Istanbul services with no explanation.

### Suggested fix
-always store the subscription returned by `addNotificationResponseReceivedListener` and call `subscription.remove()` (or `Notifications.removeNotificationSubscription(sub)`) in the cleanup, regardless of `navigationRef`'s state.

- when permission is denied, show a banner with a "Grant location" CTA that opens settings, and either disable the map or show a clear "Showing default region" tag. The half-built block in the file is a good starting point.

### LINKED ISSUE #554: Bug: Mobile Profile page show badges
When the user click add badges at mobile profile page, badges are not coming. Need fix.

### LINKED ISSUE #555: Bug: Mobile add offer flow is buggy
Add offer flow buggy with keyboard. Need Fix

yusufizzetmurat:
## #452 — mobile API client has no 401/refresh handler

**Real bug:** Once the access token expired, every screen 401'd with a generic `Error("…")` and the user was stuck on stale data until the app was force-killed. `mobile-client/src/utils/authRefresh.ts` only carries soft-refresh-skip helpers; the network layer never invoked refresh.

**Root cause:** `apiRequest` in `mobile-client/src/api/client.ts` propagated every non-OK status as a thrown error. There was no 401 detection, no refresh attempt, and no logout dispatch.

**Fix:**
- On 401, call `/auth/refresh/` once with the stored refresh token and retry the original request with the new access token.
- Concurrent 401s share a single in-flight refresh promise so N parallel requests trigger one refresh, not N.
- Refresh / login / logout endpoints bypass the cascade so they cannot recurse.
- When refresh has no token, returns 4xx, or no refresh token exists, the client clears in-memory + persisted tokens and emits `auth:logout` via a tiny in-module event bus. `RootNavigator.tsx` subscribes to that event and hands off to `AuthContext.logout()`, which resets user state — the navigator falls back to the unauthenticated tree.

**Test:** `mobile-client/src/api/__tests__/client_401.test.ts` covers (a) 401 → refresh ok → retry succeeds, (b) 401 → refresh fails → tokens cleared and `auth:logout` emitted, (c) no recurse on `/auth/refresh/`, (d) no refresh attempt without a refresh token, (e) non-401 errors propagate unchanged.

**Verification:** All 21 API test suites green (163 tests).

## #453a — push-notification listener leak

**Real bug:** A single push tap fired multiple navigations, one per stacked listener.

**Root cause:** The response-listener effect in `usePushNotifications.ts` declared `navigationRef` as a dep. When React Navigation handed back a fresh `navigationRef` (cold-start init, hot reload, deep-link mount), the cleanup → re-attach cycle ran and the previous subscription was sometimes still alive when the new one attached.

**Fix:**
- Mirror `navigationRef` into a holder ref so the effect can drop it from the dep array without losing the latest navigation target — the listener pulls the holder ref's current value at fire time.
- Both listener effects defensively call `subscription.remove()` on the ref before attaching a new one, so a stale subscription from a hot reload cannot coexist with a fresh one.

**Test:** `mobile-client/src/__tests__/components/usePushNotifications.test.tsx` covers (a) one listener attached at mount, removed at unmount, (b) re-render with a different `navigationRef` keeps at most one listener alive at any point, (c) a tap is routed through the latest `navigationRef`, not the one captured at mount.

**Verification:** Hook tests pass; full suite still green.

## #453b — silent map fallback

**Real bug:** When the Mapbox token was missing or the network was offline, `MapScreen.tsx` mounted an empty WebView and the user saw a featureless surface with no explanation.

**Root cause:** No fallback UI in either branch (token-missing or asset-load-failed).

**Fix:** Render a labelled `MapUnavailable` placeholder with copy ("Map unavailable — check your connection or configuration") and a Retry CTA when `getMapboxToken()` returns nothing or the WebView/asset load fails. Retry remounts the WebView via a key bump and re-runs the asset-load effect so a transient outage clears with one tap.

**Test:** `mobile-client/src/__tests__/components/MapScreenFallback.test.tsx` asserts the placeholder + Retry CTA render when the token is missing and the WebView itself is not mounted in that branch.

## #555 — add-offer keyboard misbehaviour

**Real bug:** Typing into the offer-form inputs scrolled the focused input UP, partially behind the keyboard, instead of out from under it.

**Root cause:** `ServiceWizard.tsx` set `keyboardVerticalOffset={Platform.OS === "ios" ? 90 : 0}`, but the wizard always mounts inside a `SafeAreaView` + custom `topBar` (`PostOfferScreen`, `PostNeedScreen`, `PostEventScreen`). The KAV's reference frame already sat below those — the hardcoded 90 was double-counting the header.

**Fix:** Set `keyboardVerticalOffset` to 0; the wrapping screen already paid for the safe area top + topBar height.

**Verification:** Manual: focus title/description on Create Offer, the field stays above the keyboard. Build still type-checks via `npx tsc --noEmit`.

## #554 — profile add-badges has no entry point

**Real bug:** The compact badge showcase rendered nothing when the user had no featured badges, so there was no discoverable path from the profile to the picker (which lives under Settings → Edit profile → Showcase).

**Root cause:** `BadgeShowcase` (compact variant) returned `null` for the empty case and `ProfileHero` had no fallback CTA.

**Fix:** When `mode === "own"` and the badge list is empty, render a small "Showcase a badge" pill in the badge slot. `ProfileScreen` wires it to `navigation.navigate("ProfileEdit", { initialTab: "showcase" })`, which loads the picker grid via `getAchievementProgress()` with one tap.

**Verification:** Manual: profile with no featured badges shows the new pill; tapping it lands on the Showcase tab with badges populated.

## Suite

`cd mobile-client && npx tsc --noEmit && npm test -- --ci` — 183 tests pass across 28 suites (was 173 across 25 before this PR).

Closes #452. Closes #453. Closes #554. Closes #555.

### DIFF
--- a/mobile-client/src/__tests__/components/MapScreenFallback.test.tsx
+++ b/mobile-client/src/__tests__/components/MapScreenFallback.test.tsx
@@ -1,0 +1,95 @@
+/**
+ * Issue #453b: when the Mapbox token is missing or the HTML asset fails to
+ * load, the map screen used to render an empty WebView — a featureless,
+ * unexplained surface. Now it renders a labelled placeholder with a Retry CTA.
+ */
+
+import React from 'react'
+import { render, fireEvent } from '@testing-library/react-native'
+
+// Mock the env module BEFORE importing the screen so getMapboxToken() returns
+// undefined for the no-token test path.
+jest.mock('../../constants/env', () => ({
+  getMapboxToken: jest.fn(() => undefined),
+  getApiUrl: () => 'http://localhost/api',
+  normalizeRuntimeUrl: (v: string | null | undefined) => v,
+}))
+
+jest.mock('@react-navigation/native', () => ({
+  useNavigation: () => ({
+    canGoBack: () => false,
+    goBack: jest.fn(),
+    navigate: jest.fn(),
+  }),
+}))
+
+jest.mock('react-native-safe-area-context', () => ({
+  useSafeAreaInsets: () => ({ top: 0, bottom: 0, left: 0, right: 0 }),
+  SafeAreaView: ({ children }: { children: React.ReactNode }) => children,
+}))
+
+jest.mock('@expo/vector-icons/Ionicons', () => 'Ionicons')
+jest.mock('@expo/vector-icons', () => ({ Ionicons: 'Ionicons' }))
+
+jest.mock('expo-asset', () => ({
+  Asset: { fromModule: () => ({ downloadAsync: jest.fn(), localUri: '' }) },
+}))
+
+jest.mock('expo-file-system', () => ({
+  File: jest.fn().mockImplementation(() => ({
+    text: jest.fn().mockResolvedValue('<html></html>'),
+  })),
+}))
+
+jest.mock('expo-location', () => ({
+  __esModule: true,
+  getForegroundPermissionsAsync: jest.fn().mockResolvedValue({ granted: false }),
+  requestForegroundPermissionsAsync: jest.fn().mockResolvedValue({ granted: false }),
+  getCurrentPositionAsync: jest.fn(),
+  Accuracy: { Balanced: 0 },
+}))
+
+jest.mock('react-native-webview', () => {
+  const React = require('react')
+  const { View } = require('react-native')
+  return {
+    __esModule: true,
+    default: React.forwardRef(({ testID }: { testID?: string }, _ref: unknown) =>
+      React.createElement(View, { testID: testID ?? 'mocked-webview' }),
+    ),
+  }
+})
+
+jest.mock('@react-native-community/slider', () => 'Slider')
+
+jest.mock('../../api/services', () => ({
+  listServices: jest.fn().mockResolvedValue({ results: [] }),
+}))
+
+import MapScreen from '../../presentation/screens/MapScreen'
+import * as env from '../../constants/env'
+
+describe('MapScreen fallback', () => {
+  beforeEach(() => {
+    ;(env.getMapboxToken as jest.Mock).mockReset()
+  })
+
+  it('shows the unavailable placeholder when no Mapbox token is configured', () => {
+    ;(env.getMapboxToken as jest.Mock).mockReturnValue(undefined)
+    const { getByText, getByLabelText, queryByTestId } = render(<MapScreen />)
+    expect(getByText('Map unavailable')).toBeTruthy()
+    expect(getByLabelText('Retry loading the map')).toBeTruthy()
+    // The WebView itself should not be on screen in the placeholder branch.
+    expect(queryByTestId('mocked-webview')).toBeNull()
+  })
+
+  it('exposes a retry CTA the user can tap', () => {
+    ;(env.getMapboxToken as jest.Mock).mockReturnValue(undefined)
+    const { getByLabelText } = render(<MapScreen />)
+    const retry = getByLabelText('Retry loading the map')
+    // No throw on press: the handler is wired and idempotent — re-running
+    // the fallback branch is safe because the token is still missing.
+    fireEvent.press(retry)
+    expect(getByLabelText('Retry loading the map')).toBeTruthy()
+  })
+})
--- a/mobile-client/src/__tests__/components/usePushNotifications.test.tsx
+++ b/mobile-client/src/__tests__/components/usePushNotifications.test.tsx
@@ -1,0 +1,193 @@
+/**
+ * Issue #453a: the notification-response listener stacked on every
+ * navigationRef identity change. A single push then fired N navigations.
+ *
+ * Regression coverage: re-render the host component with a new navigationRef
+ * and assert that exactly one listener is alive at any point — the previous
+ * subscription must be removed before the new one attaches.
+ */
+
+import React from 'react'
+import { render, act } from '@testing-library/react-native'
+import * as Notifications from 'expo-notifications'
+
+import { usePushNotifications } from '../../hooks/usePushNotifications'
+
+jest.mock('expo-notifications', () => ({
+  __esModule: true,
+  setNotificationHandler: jest.fn(),
+  getPermissionsAsync: jest.fn().mockResolvedValue({ status: 'granted' }),
+  requestPermissionsAsync: jest.fn().mockResolvedValue({ status: 'granted' }),
+  getExpoPushTokenAsync: jest.fn().mockResolvedValue({ data: 'tok' }),
+  getLastNotificationResponseAsync: jest.fn().mockResolvedValue(null),
+  dismissAllNotificationsAsync: jest.fn().mockResolvedValue(undefined),
+  setNotificationChannelAsync: jest.fn().mockResolvedValue(undefined),
+  addNotificationReceivedListener: jest.fn(() => ({ remove: jest.fn() })),
+  addNotificationResponseReceivedListener: jest.fn(() => ({ remove: jest.fn() })),
+  AndroidImportance: { MAX: 5 },
+}))
+
+jest.mock('../../context/AuthContext', () => ({
+  useAuth: () => ({ isAuthenticated: true }),
+}))
+
+jest.mock('../../store/useNotificationStore', () => ({
+  useNotificationStore: {
+    getState: () => ({ fetchUnreadCount: jest.fn() }),
+  },
+}))
+
+jest.mock('../../store/useToastStore', () => ({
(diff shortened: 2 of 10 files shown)


sgunes16:
@diclenaz7 need your comments on here

PR #574 — Manual test plan

  Mobile-only PR. Run on a physical device or simulator with the dev backend reachable (default
  http://localhost:8000). All five scenarios are independent — pick any subset; nothing depends on a
  previous step.
  
  Setup

  # Terminal 1 — backend
  make dev

  # Terminal 2 — mobile
  cd mobile-client
  npm install
  npx expo start --clear

  Open the app on iOS Simulator (i) or Android Emulator (a). Register a fresh test account or log in with
   an existing one — note the email so you can re-login later.

  Scenario 1 — Issue #452 — auth lifecycle on 401

  Goal: a 401 in the wild triggers exactly one refresh, retries the original request, and on
  unrecoverable failure logs the user out exactly once across all concurrent requests.

  1a — happy refresh-and-retry

  1. Log in.
  2. In Django shell or admin, shorten the access-token lifetime so it expires within ~30s, or wait until
   the configured access-token TTL passes naturally.
  3. Trigger a screen that does several API calls at once (e.g. open the home feed → it fetches services
  + user + notifications in parallel).

  Pass criteria:
  - No banner / red toast.
  - Backend logs show one POST /api/auth/refresh/.
  - All parallel calls resolve with their normal payload.
  - The user stays on the same screen, no logout flash.

  1b — unrecoverable refresh failure → single logout

  1. Log in, then quit the app.
  2. In your device's secure storage / AsyncStorage debugger (or via Expo Dev Tools → Storage), corrupt
  the stored refresh_token (overwrite it with bad-refresh). Alternatively, on the backend, blacklist the
  user's refresh token via Django admin.
  3. Reopen the app. The session is loaded → first authenticated screen 401s.
  4. Open three tabs/screens in quick succession to fan out 401s.

  Pass criteria:
  - App returns to the auth/login stack.
  - Console (or Sentry) shows exactly one auth:logout emission.
  - AsyncStorage access_token and refresh_token are both cleared.
  - No double navigation transitions, no "logout" toast firing twice.

  1c — refresh endpoint must not recurse

  1. Manually point the device at a backend that returns 401 for POST /api/auth/refresh/ itself.
  2. Open any screen that triggers an API call.

  Pass criteria:
  - Single 401 from the original endpoint, single 401 from /auth/refresh/, no infinite loop.
  - App lands on the login screen.
  
  Scenario 2 — Issue #453a — push notification listener does not stack

  Goal: tapping a single push notification fires exactly one navigation, even after hot reloads or
  navigationRef swaps.

  1. With the app logged in, send yourself a push (use Expo's "Push Notifications Tool" with your
  device's Expo push token, payload pointing at any in-app deep link, e.g. a service detail).
  2. Tap the notification from the lock screen / banner.
  3. Pass: lands on the target screen exactly once, no double-pushes in the navigation stack.
  4. Hot-reload the app (r in Metro), repeat steps 1–2.
  5. Pass: still exactly one navigation. Before the fix, hot-reload re-attached the listener and you'd
  see two pushes.
  6. Repeat after backgrounding then foregrounding the app a few times — still one navigation per tap.

  Scenario 3 — Issue #453b — map fallback UI

  Goal: when Mapbox is unavailable, the user sees a labelled placeholder + Retry, not a blank surface.

  3a — token missing

  1. In mobile-client/.env, comment out or empty EXPO_PUBLIC_MAPBOX_TOKEN.
  2. npx expo start --clear.
  3. Open the app, navigate to the Map screen.

  Pass criteria:
  - Visible copy "Map unavailable — check your connection or configuration".
  - A Retry button is rendered.
  - No empty WebView, no native crash.

  3b — recovery via Retry

  1. Restore the token in .env, without restarting the app — emulate a transient network/config recovery.
  2. Tap Retry.

  Pass criteria:
  - WebView remounts (key bump) and the asset-load effect re-runs.
  - Tiles render. (If the env var is bundled at build time, you may need a reload — the regression that
  matters is the placeholder + Retry rendering at all.)
  
  Scenario 4 — Issue #555 — Add-offer keyboard offset

  Goal: focused inputs on the offer/need/event wizard stay above the keyboard, not partially behind it.

  1. Log in, tap + / Post Offer (or Post Need / Post Event).
  2. Tap the Title input. Keyboard opens.
  3. Pass: the entire title field is visible above the keyboard.
  4. Tap the Description input (further down the form).
  5. Pass: the field scrolls into view and sits above the keyboard, not under or partially clipped by it.
  6. Repeat on Post Need and Post Event wizards.
  7. Test on both iOS (notch device + non-notch) and Android.

  Before the fix, the field was pushed up by 90 pt extra and slid behind the topBar / safe area instead
  of out from under the keyboard.

  Scenario 5 — Issue #554 — Profile add-badges entry point

  Goal: a user with no featured badges has a discoverable path to the badge picker.

  1. Log in as a user who has zero featured badges (a fresh account is the easiest).
  2. Open Profile.

  Pass criteria:
  - The badge slot in the profile hero shows a small "Showcase a badge" pill (instead of being empty).
  - Tap the pill → navigates to Profile Edit → Showcase tab.
  - The picker grid loads and is populated by getAchievementProgress().
  - Selecting a badge and saving brings the user back to the profile with the badge now featured (and the
   "Showcase a badge" pill replaced by the actual badge).
   
  3. Open the same profile from another user's device → the pill is not visible (mode === "own" only).

  Smoke test (after all five)

  cd mobile-client
  npx tsc --noEmit
  npm test -- --ci

  Expect 183 tests / 28 suites green. If the suite is red after a manual change, revert and retry — the
  unit suite is the cheaper trip-wire.

  Rollback signals — fail any of these and stop

  - Any path through Scenario 1 that produces two or more auth:logout emissions for one logical session
  end.
  - Any infinite refresh loop (/auth/refresh/ 401 → retry → /auth/refresh/ 401).
  - Push tap navigating twice after a hot reload.
  - Map screen rendering blank with no Retry option.
  - Wizard input visibly clipped by the keyboard on either platform.
  - "Showcase a badge" pill not appearing on a zero-badge own profile.

  If anything fails, capture the screen + Metro logs and post them on the PR — single-flight bugs
  especially are easy to miss without logs.

sgunes16 (approved):

PR #613, opened 2026-05-10
fix(frontend): wire chip diversification into the live Browse grid

yusufizzetmurat:
## Summary

Started as a one-line wiring fix for `diversifyByChip` on the Browse grid. Manual testing surfaced a chain of related bugs across the chip-identity / ranking / mobile-map stack, all addressed here. Several of the fixes informed each other (the diversifier needed `pillIdentity` to be the source of truth for the renderer too, the mobile sort had to match the web's `composite_score` for chip parity to be observable on mobile), so they ride together rather than being split into PRs that would each block on the others.

### Frontend — chip diversification on the Browse grid
- Wire `diversifyByChip` into `displayServices` in `DashboardPage`. Identity is now `pillIdentity` (mirrors `SmartPill`'s priority chain: signal chip → newcomer → capacity → explore_pool), so the diversifier sees the same identities the user sees.
- Bump `diversifyByChip` lookahead 4 → 7. With dense seeds (~80% follow-saturated) the previous window only reached the first off-chip card and left positions 3-5 as a monochrome cluster. 7 spans both visible rows above the fold.
- Promote a competitive interest signal (tag/cooccur/engagement) over a saturated `follow` win when the interest axis is at least half as strong (`FOLLOW_PROMOTION_RATIO = 0.5`). Without this the dense follow seed labels almost every card "From your network".
- Indirect-follow + newcomer-owner cards render the "Rising newcomer" pill instead of "From your network" (raw follow signal < 1.0). Direct follows keep "From your network" because that *is* the actual discovery story.
- Diversifier partitioned around the inactive (denied/cancelled) tail so the swap window cannot pull a denied row above active ones.
- Skip diversifier swap when the previous identity is `'none'` — those runs are user-invisible, perturbing ranking to break them just reorders the page for no benefit (Copilot review feedback).

### Backend — ranking
- Online services no longer get a free top-of-feed slot just because the viewer has location enabled. PostGIS `Distance(location, viewer_point)` returned `NULL` for online listings, which collapsed `composite_score` to `NULL`, which Postgres `ORDER BY DESC` sorts NULLS FIRST. `Coalesce(distance, 0.0)` so online services rank against their `hot_score * social_boost` like everyone else.
- Test-isolation fix for `test_register_two_keys_for_same_user_keeps_both` under pytest-xdist + shared Redis.

### Mobile
- HomeScreen feed sort defaulted to `latest`, so the freshest demo card sat at position 1 on mobile while the same viewer's web feed correctly led with the engagement-weighted top result. Add `sort: "hot"` so mobile and web see the same order.
- Mapbox WebView: inline the HTML as a TS constant so `Asset.downloadAsync` can't serve a stale copy across dev rebuilds; flip `mapReady` on either `loaded` or `ready` to break a bridge deadlock; prefer `pendingServices` over the stale empty initial array when `map.on("load")` fires after `updateServices` has landed.
- Mapbox markers paint after a delayed `updateServices` even when the map already finished loading.
- Don't forward the distance filter unless the user opens the range slider — the simulator's San Francisco default location was hiding every Istanbul listing.
- Align mobile map marker palette with web theme tokens (GREEN / BLUE / AMBER) so the dot colour matches the type pill on the card.
- Drop the Mapbox `NavigationControl` on mobile — pinch-to-zoom is the native gesture, the `+/-` buttons collided with the find-me FAB.
- Refetch the dashboard tag-chip strip when the authenticated user identity changes. HomeScreen is reachable without auth (Login lives inside ProfileStack), so the first chip fetch landed before the token was installed and `/featured/chips/` served its anonymous global-tag fallback. Empty `useEffect` deps pinned that response forever — after login the user kept seeing global tags instead of the personalized chips the web Browse grid surfaces.

## Test plan

- [x] `frontend`: `npm test -- src/test/pages/DashboardPage.test.tsx --run`
- [x] `frontend`: `npm test -- src/test/components/ForYouCarousel.chip.test.tsx --run`
- [x] `mobile-client`: `npx jest && npx tsc --noEmit`
- [x] `backend`: existing pytest suite + new `test_cache_utils` isolation
- [x] Manual: Browse grid on Elif's demo account no longer renders monochrome chip rows
- [x] Manual: online listings rank by `hot_score`, not by being `NULL`-distance
- [x] Manual: mobile map renders Istanbul markers with web-matching colours, no zoom-button collision, chip strip reflects logged-in identity

## Out of scope (follow-up PRs)

- Map search dropdown + signal-chip filters → PR #615

### DIFF
--- a/backend/api/tests/integration/test_service_hot_social_boost.py
+++ b/backend/api/tests/integration/test_service_hot_social_boost.py
@@ -101,3 +101,77 @@ class TestServiceHotSocialBoost:
         ordered_ids = self._service_ids_in_order(response.data)
         assert str(second_degree_service.id) == ordered_ids[0]
         assert str(disconnected_service.id) == ordered_ids[1]
+
+
+@pytest.mark.django_db
+@pytest.mark.integration
+class TestServiceHotProximityWithOnline:
+    """Online services must rank by hot_score alone when proximity is active.
+
+    Regression: the Phase 2 ranking annotates `distance = Distance(location,
+    viewer_point)`, which is NULL for Online services because they have no
+    `location` PointField. The proximity expression `1.0 / (1.0 + distance /
+    half_life)` then evaluates to NULL, propagates into `composite_score`,
+    and Postgres' default for `ORDER BY composite_score DESC` is NULLS
+    FIRST -- so every Online service piled to the top of every
+    location-aware feed regardless of its hot_score. Treat NULL distance as
+    a neutral 0 m so Online services compete on hot_score with the closest
+    in-person rows instead of jumping the queue.
+    """
+
+    def _service_ids_in_order(self, response_data):
+        results = response_data.get('results', response_data)
+        ours = [item for item in results if item['title'].startswith('[HPO]')]
+        return [item['id'] for item in ours]
+
+    def test_low_score_online_does_not_outrank_high_score_inperson_under_location(self):
+        viewer = UserFactory()
+
+        # Online row with the lowest hot_score in the matched set. Pre-fix it
+        # still landed at the top because its NULL composite_score sorted
+        # before any concrete float in `ORDER BY ... DESC NULLS FIRST`.
+        online_low = ServiceFactory(
+            user=UserFactory(),
+            type='Offer',
+            schedule_type='One-Time',
+            max_participants=1,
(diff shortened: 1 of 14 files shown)


yusufizzetmurat:
image

diclenaz7 (approved):

PR #616, opened 2026-05-10
Add useKeyboardHeight hook for dynamic keyboard height management. Up…

### LINKED ISSUE #614: Mobile App UI Flaws
There are some small ui problems in the app
- the keyboard opening input views are often blocking the input component or the submit button
- safe areas are not correctly handled for some devices
- responsiveness of the ui are broken in some screens

these should be tested across the app and should be fixed.

diclenaz7:
…dated ImagePreviewModal and ReportModal to adjust padding based on keyboard visibility. Enhanced various chat components with KeyboardAvoidingView for better user experience during text input.


yusufizzetmurat:
image

yusufizzetmurat:
image

yusufizzetmurat (changes requested):
Thanks for picking this up, going through the diff a few of them need another pass before merge, I am also adding the screenshots for chat inputs. 

  - In ChatScreen and GroupChatScreen, the old KeyboardAvoidingView is gone but listContent still has paddingBottom: 20. ChatInputBar is now absolute with bottom: keyboardHeight + 30 and minHeight: 46, so the last ~56px of the message list sits underneath the floating bar. Bumping the list's bottom padding to match the bar's footprint should fix it.
  - GroupChatScreen imports useKeyboardHeight but never calls it. Looks like a leftover from the refactor, the screen won't react to the keyboard at all in its current state.
  - PublicChatScreen ends up applying the offset twice: once on the wrapper View (paddingBottom: keyboardHeight) and again on the FlatList's contentContainerStyle (paddingBottom:
  keyboardHeight + 100). With the keyboard up that stacks to ~400px of empty space above the bar.
  - TopicDetailScreen wraps the composer in <View style={[flex, { paddingBottom: keyboardHeight }]}>, but the composer is the absolute ChatInputBar, so paddingBottom doesn't move it. You get a gap under the post list while the bar floats on its own.

  A few smaller ones while you're in there:

  - ChatInputBar takes insets into getStyles but never reads it. The hardcoded + 30 ignores safe-area, so on devices with insets.bottom = 34 the bar sits a touch too close to the home indicator.
  - useKeyboardHeight ends up with two subscriptions per screen (once in the screen, once inside ChatInputBar). Probably cleaner to pass it down as a prop, or drop the screen-level call.
  - The useKeyboardHeight() + 2 in PublicChatScreen and TopicDetailScreen could use a quick comment, or maybe roll the offset into the hook itself so callers don't have to know about it.

yusufizzetmurat:
Refs #614 .

yusufizzetmurat (approved):
Approved.

Issue #579, opened 2026-05-10
Epic: Tech Debt Cleanup and Final QA Checklist


yusufizzetmurat:
## Context
This issue tracks the final system polish, technical debt reduction, and overall stability checks required across the stack. 

### DevOps & Build
- [x] Simplify existing DevOps configuration and pipelines.
- [x] Streamline and optimize the `Makefile`.

### Refactoring & Code Quality
- [ ] Audit and remove unused libraries and dependencies.
- [ ] Purge dead code and unnecessary/stale comments.
- [ ] Deduplicate shared logic.
- [ ] Refactor core modules to reduce cyclomatic complexity (only low risk)
- [x] Update Swagger Documentation.

### UI/UX Polish
- [ ] Complete UI/UX pass on all frontend/web pages.
- [ ] Complete UI/UX pass on all mobile screens.

### Performance & Backend
- [ ] Run performance profiling and apply necessary optimizations.
- [ ] Perform scalability checks on critical infrastructure.
- [ ] Audit backend stability and error handling.

### Testing & Validation
- [ ] Review existing test suite for coverage and reliability.
- [ ] Evaluate and execute mutation testing.
- [ ] Perform final end-to-end (E2E) validation to ensure all features are functioning properly.

yusufizzetmurat:
feature-9/04-fr-09d — Service QR generation: backend endpoint exists, no FE button wires it up.
feature-9/05-fr-09e — Service QR expiration: needs a time-mock harness Playwright doesn't have.
feature-8/07-fr-08g — Service QR scan: no camera-driven scanner UI exists; Playwright can't drive a real camera.

yusufizzetmurat:
**FR-13j (public discussion tab on service detail) is being dropped — please remove it from the SRS.**

Rationale:
- Backend `CommentViewSet.create` explicitly returns 405 by policy: *"Service comments are read-only. Verified reviews are created from reputation submissions for completed exchanges."* Reviews are the only write path for service-level comments.
- Building the surface would require backend re-enable + a tabbed Reviews/Discussion UI + new moderation/throttle policy — out of scope for the milestone.
- We are not pursuing this feature; it should not stay parked under the testing/validation track of this epic either.

Action items:
- [ ] Remove FR-13j from the SRS document (and the wiki copy).
- [x] Delete the placeholder Playwright spec — done in PR #607 (commit drops `frontend/tests/e2e/feature-13/10-fr-13j.spec.ts`).

[...]

PR #621, opened 2026-05-11
Fix QR 404, #618 hot_score regression, Online proximity, follow boost

### LINKED ISSUE #618: Reputation: Offer hot_score changes during window-close batch when no evaluations were written
## What's happening

`backend/api/tests/integration/test_reputation_api.py::TestFR16eWindowCloseHotScoreContract::test_offer_hot_score_unchanged_when_no_writes_before_close` fails consistently (3/3 runs locally).

## Symptom

Open an Offer handshake, capture `service.hot_score`, expire the evaluation window via the test helper, run `process_feedback_windows`, then re-read `service.hot_score`. The assert that hot_score is unchanged fails:

\`\`\`
E       assert 0.236506 == 0.170152
\`\`\`

The close command itself reports \"No expired feedback windows found.\" so the recompute must be happening in a signal fired by the helper's \`handshake.save()\` (it touches \`evaluation_window_ends_at\` / \`evaluation_window_closed_at\`).

## Class contract

The test's class docstring is explicit:

> Service hot_score (Offer/Request): Updated at write time by Django signals (post_save on ReputationRep / NegativeRep). The batch command \`process_feedback_windows\` does NOT recalculate it.

So either:
- a) A signal is now recomputing hot_score on handshake updates that aren't reputation-related (regression), or
- b) The contract has shifted intentionally and the test is outdated.

## Status

- Test xfailed (`strict=False`, reason references this issue) so the suite stays green.
- Per #579 finalization scope, ranking behaviour stays as-is — this needs decision before any fix.

Repro: `make infra-up && cd backend && .venv/bin/python -m pytest api/tests/integration/test_reputation_api.py::TestFR16eWindowCloseHotScoreContract::test_offer_hot_score_unchanged_when_no_writes_before_close --no-cov`

yusufizzetmurat:
Four real bugs that came up while finalizing #579.

## QR Show Attendance returned "Resource not found."

The organizer's own active event was getting filtered out of `self.get_object()` on the QR endpoint, so they got a 404 toast instead of a token. `get_queryset()` narrows visibility for everything that isn't `update`/`partial_update`/`destroy`, and `generate_qr_token` / `get_qr_token` weren't in that list — so the search-engine + onboarding-skills filters could drop the organizer's own event.

Added both QR actions to the exemption tuple. Authorization stays at the service layer: `EventHandshakeService.generate_qr_token` already raises `PermissionError` for non-organizers and `ValueError` for wrong type/state/window.

Also added `api/tests/integration/test_event_qr_api.py` with 7 tests covering the ViewSet HTTP path — this whole layer was previously untested (unit tests only hit the service method directly), which is exactly why this regression shipped.

## #618 — Offer hot_score recomputed on no-op handshake save

`process_feedback_windows` uses `_expire_window()` to flip the evaluation-window timestamps. That `handshake.save(update_fields=[...])` triggered the post_save signal, which recomputed `service.hot_score` with time-sensitive components — drifting the value even though no ranking input actually changed.

Short-circuited the signal when `update_fields` is explicit and contains no hot_score-relevant fields (status / provisioned_hours). Also corrected `test_offer_hot_score_unchanged_when_no_writes_before_close` to capture the baseline from the DB after the handshake-creation signal has run; the old test was comparing a stale in-memory value, which masked the real semantics.

Closes #618.

## Online services ranked #1 once location was enabled

PostGIS `Distance()` returns NULL when the service has no location point. The previous fix `Coalesce(NULL, 0)`'d the distance — which gave Online services `proximity_factor = 1.0`, the maximum boost — so every Online row floated above every in-person row regardless of how close the in-person row was.

Switched to a `Case`: if `location IS NULL`, use 0.5 (the half-life value — proximity-neutral, neither boosted nor penalized). Also fixed the non-hot location-sorted branch in `search_filters.py` to `nulls_last=True` so Online rows don't sit at the top of distance-ordered results either.

Added a regression test under `test_proximity_ranking.py`:
`test_online_service_does_not_outrank_close_inperson_at_equal_hot_score`.

## Follow boost was only wired into sort=hot

The discovery feed defaults to "latest" (`-is_pinned, -created_at`). A service from someone you follow looked exactly like a stranger's. Added an `is_from_followed_user` Exists annotation (for authenticated viewers only) as a tiebreaker between `-is_pinned` and `-created_at`. The follow boost lives at the same priority level as pinned: pinned first, then services from people you follow, then newest.

Unxfailed `TestFollowSystem` since all six tests now pass. Note: PR #619 also touches the same `TestFollowSystem` xfail markers; whichever lands first wins, the other will need a trivial rebase.

## Verification

- Targeted suites (`test_event_qr_api`, `TestFR16eWindowCloseHotScoreContract`, `test_proximity_ranking`, `TestFollowSystem`) all pass.
- Full backend integration suite: 852 passed, 3 xfailed, 8 xpassed (the 8 XPASSes are the same stale-xfail markers PR #619 cleans up; harmless under `strict=False`).

### DIFF
--- a/backend/api/ranking_personalized.py
+++ b/backend/api/ranking_personalized.py
@@ -397,8 +397,16 @@ def score_for_you(services, viewer) -> list[tuple]:
         seconds_since = (now - last_seen) if last_seen else None
         recency = recency_penalty(seconds_since, half_life_hours)
 
+        # When the viewer enabled location, the candidate queryset annotates
+        # `proximity_factor` on each row (1.0 at the viewer's coordinates,
+        # 0.5 at the half-life distance, 0.5 flat for Online services so
+        # they stay proximity-neutral instead of out-ranking nearby
+        # in-person rows). Bake it into the hot_score input so proximity
+        # carries into the personalized blend; without this an Online
+        # service with high tag-overlap beats every nearby in-person card.
+        proximity_factor = float(getattr(svc, 'proximity_factor', 1.0))
         score, signals = blend_for_you_score(
-            hot_score=float(svc.hot_score or 0.0),
+            hot_score=float(svc.hot_score or 0.0) * proximity_factor,
             tag=tag,
             follow=follow,
             cooccur=cooccur,
--- a/backend/api/search_filters.py
+++ b/backend/api/search_filters.py
@@ -13,7 +13,7 @@ from django.contrib.gis.geos import Point
 from django.contrib.gis.measure import D
 from django.db.models import (
     Q, QuerySet, Case, When, Value, FloatField, Sum, Exists, OuterRef,
-    Subquery,
+    Subquery, F,
 )
 
 
@@ -119,9 +119,13 @@ class LocationStrategy(SearchStrategy):
                 | Q(location__distance_lte=(user_location, D(km=distance_km)))
             )
 
+        # Online services (location IS NULL) produce a NULL distance. PostgreSQL
+        # sorts NULLs FIRST under ASC by default, which used to put every Online
+        # row above every in-person row. Force NULLs to the end so the location
+        # filter actually surfaces the closest in-person services first.
         return queryset.annotate(
             distance=Distance('location', user_location)
-        ).order_by('distance')
+        ).order_by(F('distance').asc(nulls_last=True))
 
 
 def expand_tag_qids(tag_ids):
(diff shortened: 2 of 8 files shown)


yusufizzetmurat:
image

yusufizzetmurat:
image
Android mobile tested with code version and marked as attended succeeds.

sgunes16 (approved):

PR #633, opened 2026-05-11
fix(profile): paginate offers, needs, history tabs

### LINKED ISSUE #629: Fix: Paginaiton for profile page
Pagination is implemented for mobile but it is need to web also. Need fix.

Image

yusufizzetmurat:
Closes #629.

The web profile dumped every loaded record into a single grid (page_size 50 on services, no cap on history), so a power-user portfolio meant a long single scroll. Mobile already caps each list. Reveal in pages of six on web with a "Show more (N)" button beneath each grid.

## Changes

- `frontend/src/pages/UserProfile.tsx`: add `PROFILE_PAGE_SIZE = 6`, three `visible*` counters, and a reset effect on `userId` change. Slice Offers / Needs / History before rendering and append a "Show more" button while remaining items exist. Tab counters (`offersTab.length`, etc.) still reflect totals so the chips don't lie.
- Events tab unchanged: it splits into "Created" / "Joined" sub-sections that are already short in practice; not worth a counter per sub-list yet.
- Reviews unchanged: already grouped per role with an accordion on the Organizer side.

## Test plan

- [x] `npx tsc -b` clean
- [x] `npx eslint src/pages/UserProfile.tsx` clean
- [x] `npx vitest run` — 773 tests pass
- [ ] Open `/profile` for a user with >6 offers, verify "Show more (N)" appears, decrements as you click, hides at the tail
- [ ] Switch profiles via `/public-profile/:id` and confirm the counter resets

### DIFF
--- a/frontend/src/hooks/useReveal.ts
+++ b/frontend/src/hooks/useReveal.ts
@@ -1,0 +1,23 @@
+import { useCallback, useState } from 'react'
+
+export const PROFILE_PAGE_SIZE = 6
+
+/**
+ * Tracks how many items of a long list are currently revealed. The counter
+ * resets whenever `resetKey` changes so a viewer who scrolled deep on one
+ * profile doesn't land halfway into the next.
+ *
+ * Uses the "adjust state while rendering" pattern from the React docs to
+ * reset on key change without an extra effect or ref:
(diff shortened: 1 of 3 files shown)


mzyavuz (approved):
Pagination has working as supposed to be now, approving it

PR #637, opened 2026-05-12
fix: showcase sort-mode label and onboarding redirect + dropdown (#635, #636)

### LINKED ISSUE #635: Onboarding screen errors
Onboarded users can open onboarding screen, it should be redirect to profile page 

Also location screen is buggy

Image

### LINKED ISSUE #636: Recommended Showcase is Not Showing The True Message at the top
Image

Altough ranking is not based on filter by descending clearly shown below the top message is left from an older implementation and not adressed.

yusufizzetmurat:
Closes #636.
Showcase panel kept saying the list is sorted by created_at, but the dashboard actually calls the API with sort=hot. Fixed the mapping so the panel reads composite. Tests updated to keep chronological coverage via 'newest'.

Closes #635.
Onboarding: location dropdown on step 2 was being clipped by the card. Widened the overflow exception. Also added a useEffect so already onboarded users bounce to /dashboard instead of seeing the wizard again.

### DIFF
--- a/backend/api/ranking_debug.py
+++ b/backend/api/ranking_debug.py
@@ -218,7 +218,7 @@ def _sort_mode_for(active_filter: str) -> str:
     """Map the frontend ranking mode to the backend's actual sort behaviour.
 
     See ServiceViewSet.get_queryset (views.py around line 2435):
-      - sort='hot' (only 'nearby' sets this) -> composite_score order
+      - sort='hot' (the 'nearby' filter and the dashboard 'all' filter) -> composite_score order
       - explore_only feed (only 'discovery') -> Phase 3 rotation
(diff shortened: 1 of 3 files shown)


uskudarli:
You can proceed it


I will review your process :)

sgunes16 (approved):

mzyavuz (approved):
Tested all three fixes locally.

  - Showcase sort label: "Sort & position" in the debug panel now correctly
    shows `composite_score (hot sort)` for the dashboard's default All view.
  - Onboarding redirect: navigating to /onboarding as an already-onboarded
    user correctly bounces to /dashboard.
  - Dropdown clipping: location dropdown on step 2 is fully visible.

Approved!

